PR #69, opened 2018-04-19
Custom values should be implicitly allowed if a value listener is assigned

mstahv:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/69)

### DIFF
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -77,6 +77,31 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>> implements
 
     private List<T> temporaryFilteredItems;
 
+    private int customValueListenersCount;
+
+    private class CustomValueRegistraton implements Registration {
+
+        private Registration delegate;
+
+        private CustomValueRegistraton(Registration delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public void remove() {
+            if (delegate != null) {
+                delegate.remove();
+                customValueListenersCount--;
+
+                if (customValueListenersCount == 0) {
+                    setAllowCustomValue(false);
+                }
+                delegate = null;
+            }
+        }
+
+    }
+
     /**
      * Default constructor. Creates an empty combo box.
      *
@@ -457,10 +482,22 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>> implements
     }
 
     /**
-     * Adds a listener for CustomValueSetEvent which is fired when user
-     * types in a value that don't already exist in the ComboBox.
+     * Adds a listener for CustomValueSetEvent which is fired when user types in
+     * a value that don't already exist in the ComboBox.
      *
-     * <p>As a side effect makes the ComboBox allow custom values.</p>
+     * <p>
+     * As a side effect makes the ComboBox allow custom values. If you don't
+     * want to allow a user to add new values to the list once the listener is
+     * added please disable it explicitly via the
+     * {@link #setAllowCustomValue(boolean)} method.
+     * </p>
+     *
+     * <p>
+     * The custom value becomes disallowed automatically once the last custom
+     * value set listener is removed.
+     * </p>
+     *
+     * @see #setAllowCustomValue(boolean)
      *
      * @param listener
      *            the listener to be notified when a new value is filled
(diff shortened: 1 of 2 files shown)


denis-anisimov:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 462 at r1](https://beta.reviewable.io/reviews/vaadin/vaadin-combo-box-flow/69#-LAWiTcR9yGkbCAnIAU5:-LAWiTcS8Ym2bninuTSF:bdc1ah6) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/479ade845366c2efcc5d6c507b18361a4177a8a1/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L462)):*
> ```
> setAllowCustomValue(true);
> ```

It's far away from obvious that `addCustomValueSetListener` also allows custom values.
Should be mentioned in the javadoc here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/vaadin/vaadin-combo-box-flow/69)*

mstahv:
@denis-anisimov if a user adds a handler for custom values it is quite obvious he wants to allow them. I think we actually wouldn't need the flag at all to control whether the custom values are allowed or not.

Anyways, I wrote a minimalistic documentation for the addBlaBlaListener method as it didn't actually contain anything reasonable (only non-helpful documentation copied from JS docs).

denis-anisimov:
>@denis-anisimov if a user adds a handler for custom values it is quite obvious he wants to allow them. 

Yes and no.
Yes: he wants to use it.
No: it doesn't mean that he wants to have enabled custom values _always_.

The usecase is:
- create combobox.
- it's supposed to be used in various modes.
- Register all related listeners once in the initialization phase.
- Now switch the combobox mode depending on other conditions.

E.g. I may want to allow enter custom values depending on the other UI controls.
And may be even disable it by default. It should be enabled _only_ when some UI control allows this (e.g. some checkbox is checked).

So :  yes, in most cases if I register the listener then most likely I want to have custom values enabled.
But there might be cases when this is not true.

Another observation: this approach allows to mutate the same state (`AllowCustomValue`) via two totally different and unrelated methods. This is always bad. Since it's not obvious.
There should be only one mutation method for one state.
So if the method `setAllowCustomValue` had not existed  then it would have been fine.
But as I have shown above it's actually needed.

Anyway: I remember that we have similar functionality in FW8 where it had sense for me.
So I'm not against the approach but it's not that indisputable.

denis-anisimov:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 464 at r2](https://beta.reviewable.io/reviews/vaadin/vaadin-combo-box-flow/69#-LAcrrv5DnzbPSxzs-F6:-LAcrrv5DnzbPSxzs-F7:b-pj1q2) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/2b5e7bc10e88bb5ab9c01cef82d8f7a5e38fb452/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L464)):*
> ```Java
>      *
>      * <p>As a side effect makes the ComboBox allow custom values.</p>
>      *
> ```

Should also have a reference to `setAllowCustomValue`.

Also it should mention that if you want to disable (for a while may be) custom values then call `setAllowCustomValue(false)` explicitly.

Another question (as an extra improvement): should the last listener removal switch off `allowCustomValue` ?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/vaadin/vaadin-combo-box-flow/69)*

mstahv:
setAllowCustomValue method name is also broken, that should also be fixed.

I have no idea how the combobox behaves when the flag is on but there is no listener, so I can't say how it should work or does it matter. But in approximately 99.69% 🧐 of the cases, it doesn't matter. Users either have the custom value handler or not (and using it just as a replacement of a drop-down list because of reason I don't know).

Could somebody take this to the finish, I'd have my on job to do ;-)

gilberto-torrezan:
:lgtm:

---

Reviewed 1 of 2 files at r1, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/69#-:-LC45svr0ImUpUQN0b25:bnfp4nl)*

gilberto-torrezan:
:lgtm:

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/69#-:-LCDsBy97ekgNE7lJbBX:bnfp4nl)*

PR #99, opened 2018-06-29
Request re-rendering of the items.

### LINKED ISSUE #97: Calling setRenderer with ComponentRenderer after setItems renders nothing
This code works:

```java
ComboBox<String> combo = new ComboBox<>();
combo.setRenderer(new ComponentRenderer<>((label) -> new Label(label)));
combo.setItems("one", "two");
```

This code does not (elements not rendered):

```java
ComboBox<String> combo = new ComboBox<>();
combo.setItems("one", "two");
combo.setRenderer(new ComponentRenderer<>((label) -> new Label(label)));
```

We think vaadin should handle this more gracefully.

caalador:
Fixes #97

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,6 +90,12 @@
             <version>${flow.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-components-testbench</artifactId>
+            <version>1.0.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <profiles>
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -71,7 +71,7 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
     private final KeyMapper<T> keyMapper = new KeyMapper<>();
     private Element template;
 
-    private ArrayList<T> itemsFromDataProvider = new ArrayList(0);
+    private transient List<T> itemsFromDataProvider = Collections.emptyList();
     private Registration rendererRegistration;
     private boolean refreshScheduled;
 
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -205,12 +205,10 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         rendering.getDataGenerator().ifPresent(
                 generator -> rendererRegistration = dataGenerator
                         .addDataGenerator(generator));
-        if (refreshScheduled) {
-            // We have changed the renderer, but already run the rendering.
-            // Request re-render for all items.
-            refreshScheduled = false;
-            refresh();
-        }
+
+        // We have changed the renderer, but already run the rendering.
+        // Request re-render for all items.
+        refresh(true);
     }
 
     private void unregister(Registration registration) {
@@ -579,6 +577,11 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
     private void refreshDataProvider() {
         itemsFromDataProvider = dataProvider.fetch(new Query<>())
                 .collect(Collectors.toList());
+        refresh();
+    }
+
+    private void refresh(boolean refresh) {
+        refreshScheduled = !refresh;
         refresh();
     }
 
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -104,6 +104,27 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         }
     }
 
+    private class RefreshJob implements SerializableConsumer<UI> {
+
+        @Override
+        public void accept(UI ui) {
+            if (refreshJob != this) {
+                return;
+            }
+            T value = getValue();
+            keyMapper.removeAll();
+            JsonArray array = generateJson(itemsFromDataProvider.stream());
+            setItems(array);
+            /*
+             * The value is stored as a JsonObject with a key in keyMapper.
+             * Since keyMapper is reset we have to regenerate the JsonObject
+             * based on new keyMapper.
+             */
+            setValue(value);
+            refreshJob = null;
+        }
+    }
+
     /**
      * Default constructor. Creates an empty combo box.
      */
@@ -205,6 +226,17 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         rendering.getDataGenerator()
                 .ifPresent(generator -> rendererRegistration = dataGenerator
                         .addDataGenerator(generator));
+        /*
+         * DataGenerator from above is created using renderer. The data
+         * generator participates in the generateJson method implementation. As
+         * a result data which is sent to the client depends on the renderer
+         * which is created here. It means that we need to regenerate the data
+         * from scratch once the renderer is set. So we call refresh to
+         * regenerate.
+         *
+         * This is extremely not obvious and I'm going to create a ticket about
+         * this "design".
+         */
         refresh(true);
     }
 
@@ -588,27 +620,7 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         if (refreshJob != null) {
             return;
         }
-        refreshJob = new SerializableConsumer<UI>() {
-
-            @Override
-            public void accept(UI ui) {
-                if (refreshJob != this) {
-                    return;
-                }
-                T value = getValue();
-                keyMapper.removeAll();
-                JsonArray array = generateJson(itemsFromDataProvider.stream());
-                setItems(array);
-                /*
-                 * The value is stored as a JsonObject with a key in keyMapper.
-                 * Since keyMapper is reset we have to regenerate the JsonObject
-                 * based on new keyMapper.
-                 */
-                setValue(value);
-                refreshJob = null;
-            }
-
-        };
+        refreshJob = new RefreshJob();
         runBeforeClientResponse(refreshJob);
 
     }
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -205,10 +205,6 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         rendering.getDataGenerator().ifPresent(
                 generator -> rendererRegistration = dataGenerator
                         .addDataGenerator(generator));
-
-        // We have changed the renderer, but already run the rendering.
-        // Request re-render for all items.
-        refresh(true);
     }
 
     private void unregister(Registration registration) {
@@ -544,8 +540,10 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
 
     private JsonArray generateJson(Stream<T> data) {
         JsonArray array = Json.createArray();
-        data.map(this::generateJson)
+        runBeforeClientResponse(ui -> {
+            data.map(this::generateJson)
                 .forEachOrdered(json -> array.set(array.length(), json));
+        });
         return array;
     }
 
@@ -577,11 +575,6 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
     private void refreshDataProvider() {
         itemsFromDataProvider = dataProvider.fetch(new Query<>())
                 .collect(Collectors.toList());
-        refresh();
-    }
-
-    private void refresh(boolean refresh) {
-        refreshScheduled = !refresh;
         refresh();
     }
 
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -71,11 +71,11 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
     private final KeyMapper<T> keyMapper = new KeyMapper<>();
     private Element template;
 
-    private List<T> itemsFromDataProvider = Collections.emptyList();
+    private ArrayList<T> itemsFromDataProvider = new ArrayList(0);
     private Registration rendererRegistration;
     private boolean refreshScheduled;
 
-    private List<T> temporaryFilteredItems;
+    private ArrayList<T> temporaryFilteredItems;
 
     private int customValueListenersCount;
 
@@ -103,29 +103,8 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         }
     }
 
-    private static <T> T presentationToModel(ComboBox<T> comboBox,
-            JsonValue presentation) {
-        return comboBox.getValue(presentation);
-    }
-
-    private static <T> JsonValue modelToPresentation(ComboBox<T> comboBox,
-            T model) {
-
-        if (model == null) {
-            return Json.createNull();
-        }
-        int updatedIndex = comboBox.itemsFromDataProvider.indexOf(model);
-        if (updatedIndex < 0) {
-            throw new IllegalArgumentException(
-                    "The provided value is not part of ComboBox: " + model);
-        }
-        return comboBox
-                .generateJson(comboBox.itemsFromDataProvider.get(updatedIndex));
-    }
-
     /**
      * Default constructor. Creates an empty combo box.
-     *
      */
     public ComboBox() {
         super(null, null, JsonValue.class, ComboBox::presentationToModel,
@@ -181,6 +160,26 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         setItems(items);
     }
 
+    private static <T> T presentationToModel(ComboBox<T> comboBox,
+            JsonValue presentation) {
+        return comboBox.getValue(presentation);
+    }
+
+    private static <T> JsonValue modelToPresentation(ComboBox<T> comboBox,
+            T model) {
+
+        if (model == null) {
+            return Json.createNull();
+        }
+        int updatedIndex = comboBox.itemsFromDataProvider.indexOf(model);
+        if (updatedIndex < 0) {
+            throw new IllegalArgumentException(
+                    "The provided value is not part of ComboBox: " + model);
+        }
+        return comboBox
+                .generateJson(comboBox.itemsFromDataProvider.get(updatedIndex));
+    }
+
     /**
      * Sets the TemplateRenderer responsible to render the individual items in
      * the list of possible choices of the ComboBox. It doesn't affect how the
@@ -203,9 +202,9 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
             rendering = renderer.render(getElement(), keyMapper, template);
         }
 
-        rendering.getDataGenerator().ifPresent(generator -> {
-            rendererRegistration = dataGenerator.addDataGenerator(generator);
-        });
+        rendering.getDataGenerator().ifPresent(
+                generator -> rendererRegistration = dataGenerator
+                        .addDataGenerator(generator));
         if (refreshScheduled) {
             // We have changed the renderer, but already run the rendering.
             // Request re-render for all items.
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -205,10 +205,7 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         rendering.getDataGenerator()
                 .ifPresent(generator -> rendererRegistration = dataGenerator
                         .addDataGenerator(generator));
-        if (refreshJob != null) {
-            refreshJob = null;
-            refresh();
-        }
+        refresh(true);
     }
 
     private void unregister(Registration registration) {
@@ -581,6 +578,13 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
     }
 
     private void refresh() {
+        refresh(false);
+    }
+
+    private void refresh(boolean force) {
+        if (force) {
+            refreshJob = null;
+        }
         if (refreshJob != null) {
             return;
         }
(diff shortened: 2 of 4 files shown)


denis-anisimov:
Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
Review status: all files reviewed, 2 unresolved discussions (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 74 at r2](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGOO6PM3ffnPTHIK9Yh:-LGOO6PM3ffnPTHIK9Yi:b2zhpfr) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/e1d3236825c8690130b636d819600436c3414f4c/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L74)):*
> ```
> transient 
> ```

Do you have any specific reason to remove this ?
It looks like this change doesn't relate to the issue......

As I understand this list is initialized from the data provider so there is no need to keep 
serialize it since its data may be restored from the data provider.
And that's the reason why it was `transient`.

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 211 at r2](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGOP3wWD-M6U2tE1K9e:-LGOP3wWD-M6U2tE1K9f:b-8jvl6b) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/e1d3236825c8690130b636d819600436c3414f4c/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L211)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
>         if (refreshScheduled) {
>             // We have changed the renderer, but already run the rendering.
>             // Request re-render for all items.
>             refreshScheduled = false;
>             refresh();
>         }
> ```
</details>



I would consider `refreshScheduled` as a variable scoped by the `refresh` method.
So that is should not be used outside of this method.

As a result I would suggest to modify `refresh` method so that it accepts `boolean`  parameter (overload `refresh` actually with one more method).
And it reschedule the runnable if the parameter value forces to do this.
So move this logic into `refresh`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

ZheSun88:
Review status: all files reviewed, 2 unresolved discussions (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 74 at r2](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGOO6PM3ffnPTHIK9Yh:-LGOtMAU7acFo7vhvMt-:b-bzhp3t) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/e1d3236825c8690130b636d819600436c3414f4c/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L74)):*
<details><summary><i>Previously, denis-anisimov (Denis) wrote…</i></summary><blockquote>

> ```
> transient 
> ```

Do you have any specific reason to remove this ?
It looks like this change doesn't relate to the issue......

As I understand this list is initialized from the data provider so there is no need to keep 
serialize it since its data may be restored from the data provider.
And that's the reason why it was `transient`.

</blockquote></details>

I am not sure what you mean here. It looks this is an adding not removing, and what you have explained here is just the reason why mikael adds this

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 211 at r2](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGOP3wWD-M6U2tE1K9e:-LGOtdJY7xKBtPAIBGkO:b9oemnn) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/e1d3236825c8690130b636d819600436c3414f4c/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L211)):*
<details><summary><i>Previously, denis-anisimov (Denis) wrote…</i></summary><blockquote>

> ```
>         if (refreshScheduled) {
>             // We have changed the renderer, but already run the rendering.
>             // Request re-render for all items.
>             refreshScheduled = false;
>             refresh();
>         }
> ```


I would consider `refreshScheduled` as a variable scoped by the `refresh` method.
So that is should not be used outside of this method.

As a result I would suggest to modify `refresh` method so that it accepts `boolean`  parameter (overload `refresh` actually with one more method).
And it reschedule the runnable if the parameter value forces to do this.
So move this logic into `refresh`.
</blockquote></details>

As this is a BFP, i will implement the logic based your comments, so we can process it. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 74 at r2](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGOO6PM3ffnPTHIK9Yh:-LGPRE9aAiT4HoEsQ6T-:bpczfw2) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/e1d3236825c8690130b636d819600436c3414f4c/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L74)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

I am not sure what you mean here. It looks this is an adding not removing, and what you have explained here is just the reason why mikael adds this
</blockquote></details>

Right, I'm an idiot.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

ZheSun88:
Review status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 74 at r2](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGOO6PM3ffnPTHIK9Yh:-LGPX6V29vPUGQoMGeBQ:b-6hp0g8) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/e1d3236825c8690130b636d819600436c3414f4c/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L74)):*
<details><summary><i>Previously, denis-anisimov (Denis) wrote…</i></summary><blockquote>

Right, I'm an idiot.
</blockquote></details>

Could you resolve the comment and review the changes i made for this PR? then we can get the BFP fixed. :-)

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 74 at r2](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGOO6PM3ffnPTHIK9Yh:-LGPZWyX6Kqc5AYVYzw3:b-cjzruy) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/e1d3236825c8690130b636d819600436c3414f4c/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L74)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

Could you resolve the comment and review the changes i made for this PR? then we can get the BFP fixed. :-)
</blockquote></details>

it is resolved already

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

ZheSun88:
Review status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 211 at r2](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGOP3wWD-M6U2tE1K9e:-LGP_-zU3aDDu32VQmx5:b-yq3btr) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/e1d3236825c8690130b636d819600436c3414f4c/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L211)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

As this is a BFP, i will implement the logic based your comments, so we can process it. 
</blockquote></details>

and here?

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGPZpPo79M0VtgFCF_5:byhlbhh) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
> ```Java
>     private void refresh(boolean refresh) {
>         refreshScheduled = !refresh;
>         refresh();
> ```

I don't think this is correct.
If refresh has not been scheduled then this code makes it scheduled even though there is no task which should reset the flag.

I think this method should have the implementation of the 'refresh' method with additional logic. And refresh should just delegate to refresh(false).

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed, 1 unresolved discussion (waiting on @denis-anisimov)

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

ZheSun88:
Review status: all files reviewed, 1 unresolved discussion (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGPiyxPAYCfn6iYpnsP:bmh6zu1) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
<details><summary><i>Previously, denis-anisimov (Denis) wrote…</i></summary><blockquote>

I don't think this is correct.
If refresh has not been scheduled then this code makes it scheduled even though there is no task which should reset the flag.

I think this method should have the implementation of the 'refresh' method with additional logic. And refresh should just delegate to refresh(false).
</blockquote></details>

okay, so you mean the structure should be like this?
```
private void refresh() {
    refresh(false);
}

private void refresh(boolean refresh){
refreshScheduled = refresh;
.....//same as it is now
}
```
---------------------------------
but that looks strange to me, as just from the name of the method, `refresh(false)` looks the value should not be refreshed.  however, with current implementation, `refreshScheduled = false` is the condition we check for further steps. 

The additional logic is the part which mikael added right? 

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: all files reviewed, 1 unresolved discussion (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGPsbCp67sk61DiF09n:bstvlis) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

okay, so you mean the structure should be like this?
```
private void refresh() {
    refresh(false);
}

private void refresh(boolean refresh){
refreshScheduled = refresh;
.....//same as it is now
}
```
---------------------------------
but that looks strange to me, as just from the name of the method, `refresh(false)` looks the value should not be refreshed.  however, with current implementation, `refreshScheduled = false` is the condition we check for further steps. 

The additional logic is the part which mikael added right? 
</blockquote></details>

yes for the structure.
The parameter name should be `force`.

And instead of direct assignment to the `refreshScheduled` it should not return if `refreshScheduled`  and `force`.
So the only change in the `refresh`  method which becomes now `refresh(boolean)` is: do not return if force.

Actually now I see that I don't understand why do we need force (or check `refreshScheduled`).
The implementation of `refresh()` doesn't depend anyhow in renderer.
So it means that it doesn't matter whether we have pending execution of refresh task or it's the task which has been added after the renderer had been set.

So it looks like we just need to call `refresh` in set renderer and that's it.
The refresh task will be called AFTER the renderer logic anyway.
So I don't see in `force`  parameter and any checks here.

If `refresh` depends on renderer anyhow then renderer should participate in the refresh task  scheduled via `runBeforeClientResponse`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: all files reviewed, 1 unresolved discussion (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGQBr7e3FY5GTVuqvNS:bgsfvrv) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

yes, i do agree with this (though, it doesn't work for this case and even failed with the other condition)

>So it looks like we just need to call refresh in set renderer and that's it.
The refresh task will be called AFTER the renderer logic anyway.
So I don't see in force parameter and any checks here.

But it seems there are some weird logic with component, after i moved to the necessary parts into the `runBeforeClientResponse` in `refresh`, the tests are failing.
</blockquote></details>

Which necessary parts ?
I don't see any parts at the moment which needs to be added into `runBeforeClientResponse`.
I would say that only the `refresh()` method call should be enough in the `setRenderer` method.
If not then we have to understand why. Because as I said I don't see how renderer depends on the consumer in the method.

In any case overloaded `refresh` should just improve the code in the PR.
But as I said: we should investigate why the `refresh()`call is not enough (if it is).

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

ZheSun88:
Review status: all files reviewed, 1 unresolved discussion (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGQDIFyBg5jWZiPtmNX:b-ovbkjy) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
<details><summary><i>Previously, denis-anisimov (Denis) wrote…</i></summary><blockquote>

Which necessary parts ?
I don't see any parts at the moment which needs to be added into `runBeforeClientResponse`.
I would say that only the `refresh()` method call should be enough in the `setRenderer` method.
If not then we have to understand why. Because as I said I don't see how renderer depends on the consumer in the method.

In any case overloaded `refresh` should just improve the code in the PR.
But as I said: we should investigate why the `refresh()`call is not enough (if it is).
</blockquote></details>

yes, i understand you. It looks the current fix doesn't fix the problem correctly. 

The current situation is: if we run the `runBeforeClientResponse` for only one time (that is why we have the `refreshScheduled` flag),  it will fail on this `setitems-setRender` case. 

with Mikael's fix, we basically enable the one more run for the `runBeforeClientResponse` block inside `refresh` method (one call from the `setItem`, and one call from the `setRender`) . It works, but doesn't sounds correctly. 

will investigate more.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: all files reviewed, 1 unresolved discussion (waiting on @denis-anisimov)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGQM_Px0pP_rgaawkyc:bxatemp) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

yes, i understand you. It looks the current fix doesn't fix the problem correctly. 

The current situation is: if we run the `runBeforeClientResponse` for only one time (that is why we have the `refreshScheduled` flag),  it will fail on this `setitems-setRender` case. 

with Mikael's fix, we basically enable the one more run for the `runBeforeClientResponse` block inside `refresh` method (one call from the `setItem`, and one call from the `setRender`) . It works, but doesn't sounds correctly. 

will investigate more.
</blockquote></details>

And I don't understand why it works.

It looks like there are hidden parts of renderer which somehow overlaps with refresh job.
Those parts need to be revealed and explicitly called inside refresh job.
And then renderer should just call `refresh()`.

Now it works for some unknown reason.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @denis-anisimov and @vaadin-bot)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGVr4Xh6MmrRdYJ7hQ9:b18kj3j) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
<details><summary><i>Previously, ZheSun88 (Sun Zhe) wrote…</i></summary><blockquote>

just did a long debugging, i found, when we have the code in `setItem-setRenderer` order, the `array` which we got from `generateJson` doesnt contain the render info, and that is why the component render looks empty when we check the UI.  The reason for causing this is when we change the renderer, we need to postpone the json generation until the re-rendering finishes, otherwise, the returned JsonArray will use the incorrect info for the component renderer. 

found this code here 
```
    public void setFilteredItems(Collection<T> filteredItems) {
        temporaryFilteredItems = new ArrayList<>(filteredItems);
 
        runBeforeClientResponse(ui -> {
            setFilteredItems(generateJson(temporaryFilteredItems.stream()));
            temporaryFilteredItems = null;
        });
    }
```
It looks that is using the same logic. (but i am not sure.)

</blockquote></details>

The reason of the behavior is quite easy actually.
It's the `keyMapper`.

I remember you have been already fixing similar issue.
The result of the fix was: `keyMapper.removeAll();` moved _INTO_ `runBeforeClientResponse`  task.

And similar thing here: since we are doing something with `keyMapper` via postponing this then it means that _EVERY_ access to keyMapper has to be postponed.

So if I'm right then `keyMapper` is just cleaned up _AFTER_ the `setRenderer` had done its job.
As a result `setRenderer` had used wrong data in the `keyMapper`.

So no, problem is not in the `generateJson` . The problem is `keyMapper`  usage in the setRenderer.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @denis-anisimov and @vaadin-bot)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGVvZhQBIA08w0TEB7C:b-ehit8k) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
<details><summary><i>Previously, denis-anisimov (Denis) wrote…</i></summary><blockquote>

The reason of the behavior is quite easy actually.
It's the `keyMapper`.

I remember you have been already fixing similar issue.
The result of the fix was: `keyMapper.removeAll();` moved _INTO_ `runBeforeClientResponse`  task.

And similar thing here: since we are doing something with `keyMapper` via postponing this then it means that _EVERY_ access to keyMapper has to be postponed.

So if I'm right then `keyMapper` is just cleaned up _AFTER_ the `setRenderer` had done its job.
As a result `setRenderer` had used wrong data in the `keyMapper`.

So no, problem is not in the `generateJson` . The problem is `keyMapper`  usage in the setRenderer.

</blockquote></details>

No, I'm wrong. It's not easy.

The values from `keyMapper` is  not used directly right away. It's used as function to get them on demand.
But it happens via `registerEventHandlers` method which also schedules a couple of tasks.
Apparently the order of the execution of those task matters.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

denis-anisimov:
Review status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @denis-anisimov and @vaadin-bot)

---

*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 585 at r3](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99#-LGPZpPo79M0VtgFCF_4:-LGW1PaW9matZC0V7Gic:b-g7lwmj) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/35d3bf731e9e1784e057cc41c1ec072b4e0ef9c6/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L585)):*
<details><summary><i>Previously, denis-anisimov (Denis) wrote…</i></summary><blockquote>

No, I'm wrong. It's not easy.

The values from `keyMapper` is  not used directly right away. It's used as function to get them on demand.
But it happens via `registerEventHandlers` method which also schedules a couple of tasks.
Apparently the order of the execution of those task matters.
</blockquote></details>

I'm totally wrong the `keyMapper` is not involved in the renderer at all.
The code is crappy.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/99)*

PR #143, opened 2018-10-15
Use filtering configured in DataProvider

### LINKED ISSUE #140: Issues with new lazy provider and renderers
I'm testing the recently merged lazy data provider feature, which effectively works loading data lazily but I'm experiencing some filter/rendering issues.

1. When filtering and then clearing the filter, the filtered items disappear (see demo below).
2. The filter does not always respect the server defined filter (in the demo below the server side filter is set to filter by e-mail, but the client filters by any matching text it founds in the item).

lazy-combo-box-small

Ping @pekam

pekam:
When using the overload that takes just a DataProvider, use server-side
filtering so that a filter defined in the DataProvider is used. There's
still an overload with ListDataProvider which allows client-side
filtering.

Fixes #140

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/143)

### DIFF
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -402,9 +402,6 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
 
     @Override
     public void setDataProvider(DataProvider<T, String> dataProvider) {
-        if (userProvidedFilter == UserProvidedFilter.UNDECIDED) {
-            userProvidedFilter = UserProvidedFilter.YES;
-        }
         setDataProvider(dataProvider, SerializableFunction.identity());
     }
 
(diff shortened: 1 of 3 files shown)


Legioth:
___
*[src/main/java/com/vaadin/flow/component/combobox/ComboBox.java, line 477 at r1](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/143#-LOqqp_U-DTGzQEO60_M:-LOvuzxJ9jLlYDlT8Qve:b2oqsqf) ([raw file](https://github.com/vaadin/vaadin-combo-box-flow/blob/61af2689c2c567dc0208e9cfc2a8db43748d876b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L477)):*
<details><summary><i>Previously, pekam (Pekka Maanpää) wrote…</i></summary><blockquote>

The API is the same as in V8. @legioth, do you have an explanation why it has been like this?
</blockquote></details>

The key aspect here is that the filter type is `SerializablePredicate<T>`. Whether the data is lazy loaded or in-memory isn't relatively irrelevant.

If the type signature would be `DataProvider<T, SerializablePredicate<T>>`,  then type erasure would make it conflict with the `DataProvider<T, String>` overload. At the same time, the discoverability of this API would drop significantly if `ListDataProvider` isn't explicitly mentioned.

PR #173, opened 2018-12-17
Use 'value' as item identifier to fix BFP

pekam:
Fixes vaadin/flow#4862

This is a workaround for a certain situation when client-side
value-setter is not called when the combo boxes are wrapped in a certain
PolymerTemplate structure.

Now the item.value is used as itemIdPath, which means that when setting
the selectedItem, the web component finds the value identifier
immediately from that item.

Previously it worked like so that later after setting selectedItem, the
value-property is set which fixed everything before the itemIdPath was
set. But this value-property is not going through the Polymer
property-setting in this certain scenario.

This causes some extra overhead: we need the old item.key also because
the datagenerators and renderers depend on it. So there will be
duplicate property in each data item: key and value. We need to consider if this is
worth it to fix the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/173)


pekam:
Closing this in favor of a cleaner fix.

PR #183, opened 2019-01-07
Always intialize items, even when empty

### LINKED ISSUE #169: calling setItems causes a client-side exception
`vaadin.version 12.0.1`
Setting the items to a previously-uninitialized ComboBox causes a client-side exception.
Simple demo that reproduces the issue:

```
public class MainView extends VerticalLayout {

    public MainView() {
        ComboBox<String> comboBox = new ComboBox<>();
        Button button = new Button("Click me to add items to the combobox",
                event -> comboBox.setItems("foo", "bar"));
        add(comboBox, button);
    }
}
```
Clicking the button causes a bunch of TypeErrors on the client side:
```
client-D1AD34905AC1AA5B4DBECA8FB0306D92.cache.js:197 (TypeError) : Cannot read property 'reset' of undefined
```
Workaround is to initialize the ComboBox with an empty Collection.

macprzepiora:
This fixes #169

### DIFF
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -16,11 +16,7 @@
 package com.vaadin.flow.component.combobox;
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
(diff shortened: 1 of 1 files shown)


Silver-Wulf:
For the test ComboBoxTest.setValueWithoutItems_throw, should an exception be thrown only if the dataProvider is null or if the set of items is empty in general? A fix for the test could very well be to check if the list is empty when setValue is called as well as checking for a null DataProvider.

PR #186, opened 2019-01-13
Fix: value is not updated when null in ComboBox

### LINKED ISSUE #180: Setting value to null displays a number in ComboBox
If custom values are enabled, setting value to null in value change listener causes ComboBox displays a number as value.

A code to reproduce (tested with 12.0.3, but I think the issue has been there always):

    public static class Person implements Serializable {
        private String name;
        public Person(String firstName) {
            this.name = firstName;
        }
        @Override
        public String toString() {
            return name;
        }
    }

    public MainView() {
        Person jorma = new Person("Jorma");
        Person kalle = new Person("Kalle");
        List<Person> list = Arrays.asList(jorma, kalle);

        ComboBox<Person> cb = new ComboBox<>();
        cb.setItems(list);
        cb.addValueChangeListener(e -> {
            e.getSource().setValue(null);
        });

        cb.setAllowCustomValue(true);
        add(cb); // MainView extends VerticalLayout
    }

In above MainView, just choose "Jorma" -> ComboBox shows 1, even though it should be empty.

Client side console shows this message:

    Warning: unable to determine the label for the provided `value`. Nothing to display in the text field. This usually happens when setting an initial `value` before any items are returned from the `dataProvider` callback. Consider setting `selectedItem` instead of `value`

Silver-Wulf:
This should fix input element value not being empty when setting value to null. This is because the JS selectedItemChanged method does not update value unless customInput is disabled due to selectedItem being set to null each time the input no longer matches a dropdown option.

This should fix #180 but does not address another issue that caused numbers to be displayed. I will open an issue for this.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/186)

### DIFF
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -324,6 +324,8 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
 
         if (value == null) {
             getElement().setProperty("selectedItem", null);
+            getElement().setProperty("value", "");
+            getElement().setProperty("_inputElementValue", "");
             return;
         }
 
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -459,13 +459,6 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         if (userProvidedFilter == UserProvidedFilter.UNDECIDED) {
             userProvidedFilter = UserProvidedFilter.YES;
         }
-        
-        if(getElement().getProperty("$connector") == null) {
-        	Element e = getElement();
-	        while(e.getParent() != null)
-	        	e = e.getParent();
-	        e.executeJavaScript("window.Vaadin.Flow.comboBoxConnector.initLazy($0);", getElement());
-        }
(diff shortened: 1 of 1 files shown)


Silver-Wulf:
My bad on those commits, that was supposed to be pushed to a different branch.

Silver-Wulf:
For #180, modifying the example provided by the issue
```
public static class Person implements Serializable {
        private String name;
        public Person(String firstName) {
            this.name = firstName;
        }
        @Override
        public String toString() {
            return name;
        }
    }

    public MainLayout() {
        Person jorma = new Person("Jorma");
        Person kalle = new Person("Kalle");
        List<Person> list = Arrays.asList(jorma, kalle);

        ComboBox<Person> cb = new ComboBox<>();
        cb.setItems(list);
        cb.addValueChangeListener(e -> {
            e.getSource().setValue(null);
            e.getSource().getElement().setProperty("value", "");
            e.getSource().getElement().setProperty("_inputElementValue", "");
        });

        cb.setAllowCustomValue(true);
        add(cb); // MainView extends VerticalLayout
    }
```
Running the example provided by the issue, the text field shows 1 and 2 instead of being blank when setValue(null) is called but adding the lines to set value and _inputElementValue to an empty string results in the text field being empty as intended.

Silver-Wulf:
@bogdanudrescu Could you point me in the correct direction for building/testing a vaadin environment using the local maven project I have for vaadin-combo-box?

PR #188, opened 2019-01-13
 Initialize $connector if values were not set in ComboBox element prior to page load.

### LINKED ISSUE #169: calling setItems causes a client-side exception

### LINKED ISSUE #180: Setting value to null displays a number in ComboBox

Silver-Wulf:
Initializes the $connector property on ComboBox's element if this was not done prior to changing DataCommunicator by calling the comboBoxConnector.initLazy() function.

Should fix #169

The first commit is in pull #186 and fixes #180 

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/188)

### DIFF
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -324,6 +324,8 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
 
         if (value == null) {
             getElement().setProperty("selectedItem", null);
+            getElement().setProperty("value", "");
+            getElement().setProperty("_inputElementValue", "");
             return;
         }
 
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -466,10 +466,8 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
             userProvidedFilter = UserProvidedFilter.YES;
         }
         
-        if(getElement().getProperty("$connector") == null) {
-            UI.getCurrent().getPage().executeJavaScript("window.Vaadin.Flow.comboBoxConnector.initLazy($0);", getElement());
-        }
-
+        runBeforeClientResponse(ui -> ui.getPage().executeJavaScript("window.Vaadin.Flow.comboBoxConnector.initLazy($0);", getElement()));
+        
         if (dataCommunicator == null) {
             dataCommunicator = new DataCommunicator<>(dataGenerator,
                     arrayUpdater, data -> getElement()
--- a/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -461,10 +461,10 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
         }
         
         if(getElement().getProperty("$connector") == null) {
-        	Element e = getElement();
-	        while(e.getParent() != null)
-	        	e = e.getParent();
-	        e.executeJavaScript("window.Vaadin.Flow.comboBoxConnector.initLazy($0);", getElement());
+            Element e = getElement();
+            while(e.getParent() != null)
+                e = e.getParent();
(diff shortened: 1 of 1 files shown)


Silver-Wulf:
For #169, modifying the example provided by the issue.
```
public MainLayout() {
		ComboBox<String> comboBox = new ComboBox<>();
		// comboBox.setItems(Collections.emptyList());
		Element e = comboBox.getElement();
		while (e.getParent() != null)
			e = e.getParent();
		final Element ele = e;
		Button button = new Button("Click me to run", event -> {
			if (comboBox.getElement().getProperty("$connector") == null) {
				ele.executeJavaScript("window.Vaadin.Flow.comboBoxConnector.initLazy($0);", comboBox.getElement());
			}
		});

		Button button2 = new Button("Click me to add items to the combobox", event -> comboBox.setItems("foo", "bar"));
		add(comboBox, button);
		add(comboBox, button2);
	}
```

Clicking button2 to add items causes an exception as the issue stated but pressing button prior to that causes the page to work as intended with the items being added when button2 is pressed.



For #180, modifying the example provided by the issue
```
public static class Person implements Serializable {
        private String name;
        public Person(String firstName) {
            this.name = firstName;
        }
        @Override
        public String toString() {
            return name;
        }
    }

    public MainLayout() {
        Person jorma = new Person("Jorma");
        Person kalle = new Person("Kalle");
        List<Person> list = Arrays.asList(jorma, kalle);

        ComboBox<Person> cb = new ComboBox<>();
        cb.setItems(list);
        cb.addValueChangeListener(e -> {
            e.getSource().setValue(null);
            e.getSource().getElement().setProperty("value", "");
            e.getSource().getElement().setProperty("_inputElementValue", "");
        });

        cb.setAllowCustomValue(true);
        add(cb); // MainView extends VerticalLayout
    }
```
Running the example provided by the issue, the text field shows 1 and 2 instead of being blank when setValue(null) is called but adding the lines to set value and _inputElementValue to an empty string results in the text field being empty as intended.

Silver-Wulf:
@bogdanudrescu If I am correct, the fix using runBeforeClientResponse will not prevent the errors from happening because the code will not execute before the rest of the method body in setDataProvider. I can pinpoint that the issue is related to the $connector not being initialized from the osgi resource unless the data provider is set before the page loads. Is there any way for me to run the code you posted above the moment the page is loaded?

Silver-Wulf:
So I placed the method call where the old one was. I am unsure about whether or not that works and I am wondering if that should/could have been placed within the constructor. Usually initConnector is called when the data provider is set before client connection but this should be a manual fix for cases where that does not happen. I removed the check for the $connector variable since the initLazy method has its own check.

Issue #219, opened 2019-02-20
Connector error with 2.1.1


heruan:
After upgrading to 13.0.0.beta2 (which brings 2.1.1) I get this client error from the connector:

```
Exception is thrown during JavaScript execution. Stacktrace will be dumped separately.
(TypeError) : Cannot read property 'input' of undefined
The error has occurred in the JS code: '$0, $1, $0.$connector.enableClientValidation($1)'
```

Not sure what changed, but the same exact code worked before.

heruan:
Cannot replicate anymore, so I'm closing for now; I'll reopen in case I confirm this.

heruan:
This is indeed happening with 2.1.1 but only under these conditions:

1. the component which contains the `ComboBox` is a `PolymerTemplate`
2. the component which contains the `ComboBox` has at least an `@Id` field;
3. the `ComboBox` should at least have an `asRequired()` binding.

For example:

```html
<link rel="import" href="../bower_components/vaadin-tabs/vaadin-tabs.html">

<dom-module id="combo-test">
  <template>
    <div id="navigation" part="navigation">
      <vaadin-tabs id="tabs"></vaadin-tabs>
    </div>
    <div id="content" part="content">
      <slot></slot>
    </div>

  </template>

  <script>
    class ComboTest extends Polymer.Element {
      static get is() { return 'combo-test'; }
      static get properties() {
        return {
        };
      }
    }

    window.customElements.define(ComboTest.is, ComboTest);
  </script>
</dom-module>
```

```java
@Tag("combo-test")
@HtmlImport("combo-test.html")
public class ComboTest extends PolymerTemplate<TemplateModel> implements HasComponents {

    @Id("tabs")
    protected Tabs tabs;

    public ComboTest() {
        var cb = new ComboBox<String>();
        cb.setItems("Foo", "Bar");
        var binder = new Binder<Bean>();
        binder.forField(cb).asRequired().bind(Bean::getter, Bean::setter);
        this.add(cb);
    }

    interface Bean { String getter(); void setter(String value); }
}
```

heruan:
Might have been caused by https://github.com/vaadin/vaadin-combo-box-flow/pull/202

heruan:
[...]

PR #223, opened 2019-03-01
Check whether the combobox shadow root is not null

denis-anisimov:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/223)


pleku:
This fix should only be backported to `2.1` branch, but nothing else.

denis-anisimov:
Fixes #219

PR #231, opened 2019-03-12
Don't apply new filter to DP if filter not changed

### LINKED ISSUE #227: Items not rendered partially after scrolling in large combobox
Have a look at my example based on starter project: 
[demo-combobox-items-not-rendered.zip](https://github.com/vaadin/vaadin-combo-box-flow/files/2937939/demo-combobox-items-not-rendered.zip)

The combobox has a lot of items. When scrolling down and up, some items are not rendered correctly anymore, though they still exist and can be selected:
demo-combobox-items-not-rendered

Same behavior in Vaadin 12.0.7 and 13.0.0.

pekam:
Otherwise, DataProvider will get a new instance of the filter object
-> DataCommunicator thinks filter changed
-> destroys all data
-> rendered components are destroyed
-> client still thinks that everything is fine
-> flow-component-renderer doesn't find destroyed component instances
-> components not rendered

Fix #227

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/231)

### DIFF
--- a/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -159,8 +159,6 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
             enqueue("$connector.confirm", updateId, ComboBox.this.lastFilter);
             queue.forEach(Runnable::run);
             queue.clear();
-
-            ComboBox.this.lastFilter = null;
         }
 
         private void enqueue(String name, Serializable... arguments) {
@@ -938,10 +936,11 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
 
     @ClientCallable
     private void setRequestedRange(int start, int length, String filter) {
-        this.lastFilter = filter;
-
         dataCommunicator.setRequestedRange(start, length);
-        filterSlot.accept(filter);
+        if (!Objects.equals(filter, lastFilter)) {
+            lastFilter = filter;
+            filterSlot.accept(filter);
+        }
     }
 
     @ClientCallable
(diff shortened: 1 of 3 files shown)


vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -159,8 +159,6 @@ public void commit(int updateId) {
             enqueue("$connector.confirm", updateId, ComboBox.this.lastFilter);
             queue.forEach(Runnable::run);
             queue.clear();
-
-            ComboBox.this.lastFilter = null;
```
pekam: Hey, @bogdanudrescu can you verify if it's okay to remove this line? For me it seems that it is not needed, but can you verify this as the author of this line?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -938,10 +936,11 @@ private void confirmUpdate(int id) {
 
     @ClientCallable
     private void setRequestedRange(int start, int length, String filter) {
-        this.lastFilter = filter;
-
         dataCommunicator.setRequestedRange(start, length);
-        filterSlot.accept(filter);
+        if (!Objects.equals(filter, lastFilter)) {
```
yuriy-fix: Probably it shouldn't be checked over here. Instead the logic should be changed when setting dataProvider for `filterSlot`.

pekam: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
pekam: To clarify, I need to keep the reference to the last filter, because I need to compare if the filter has changed from last request or not.

I could also create another property for this purpose, but it would be weird to have two properties saving the last filter string.

bogdanudrescu: Hi @pekam, sorry for the late reply and thanks for noticing this. I think your point makes sense and we should remove that line, always keeping the value in the `lastFilter`.

Issue #235, opened 2019-03-19
Connector problems with 2.1.2 - Cannot read property 'enableClientValidation' of undefined


lucastanzani:
I must say this may be a derivation - duplication of https://github.com/vaadin/vaadin-combo-box-flow/issues/219
My setup is vaadin 13.0.1 and combobox explicitly set to version 2.1.2 through pom dependency.
I have a form with a combobox with a binder. The underlying bound object has a @Id and  asRequired is set for the property the combobox must hold. 
Example: 

```
binder.forField(studentBox)
                .asRequired("Inserire uno studente a cui associare la partecipazione!")
                .withConverter(
                        studentService::idToStudent, studentService::studentToId,
                        "Inserire uno studente tra quelli dell'elenco")
                .bind(Participation::getStudent, Participation::setStudent);
```
The combobox displays the id of the student to add to the Participation, which is the entity.
When i call readBean(some_participation) i get this error:
client-28831013082792340517B7074033EF04.cache.js:975 Exception is thrown during JavaScript execution. Stacktrace will be dumped separately.
The error has occurred in the JS code: '$0, $1, $0.$connector.enableClientValidation($1)'

```
Uncaught TypeError: Cannot read property 'enableClientValidation' of undefined
    at Object.eval (eval at jt (client-28831013082792340517B7074033EF04.cache.js:975), <anonymous>:3:15)
    at jt (client-28831013082792340517B7074033EF04.cache.js:975)
    at it (client-28831013082792340517B7074033EF04.cache.js:931)
    at gt (client-28831013082792340517B7074033EF04.cache.js:562)
    at Pq (client-28831013082792340517B7074033EF04.cache.js:482)
    at lr.mr [as W] (client-28831013082792340517B7074033EF04.cache.js:983)
    at eA (client-28831013082792340517B7074033EF04.cache.js:866)
    at sk (client-28831013082792340517B7074033EF04.cache.js:731)
    at Dk.Ek [as W] (client-28831013082792340517B7074033EF04.cache.js:983)
    at eA (client-28831013082792340517B7074033EF04.cache.js:866)
    at Rq (client-28831013082792340517B7074033EF04.cache.js:978)
    at fr.gr [as D] (client-28831013082792340517B7074033EF04.cache.js:983)
    at zj (client-28831013082792340517B7074033EF04.cache.js:749)
    at Ej.Gj [as V] (client-28831013082792340517B7074033EF04.cache.js:983)
[...]

PR #271, opened 2019-05-29
Init $connector in ArrayUpdater.initialize and ctor

### LINKED ISSUE #270: Cannot read property '$connector' of null in latest snapshot
>>>Uncaught TypeError: Cannot read property '$connector' of null
    at Object.initLazy (comboBoxConnector.js?c659:10)
    at Object.eval (eval at qt (client-5B44EA78EA65C9AA94E2163A65E14674.cache.js:983), <anonymous>:3:38)

`initLazy` is called with null combobox

It can be reproduced editing order in bakery in dev mode

pekam:
Moved initConnector to its original position, since attach listener
caused problems. It's still needed in constructor as well to fix another
issue.

This is basically changing the code to a state as if I didn't remove the call in ArrayUpdater.initialize in #252 

fix #270

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/271)

### DIFF
--- a/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/DetachReattachPage.java
+++ b/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/DetachReattachPage.java
@@ -37,7 +37,7 @@ public class DetachReattachPage extends Div {
             add(comboBox);
             remove(comboBox);
         });
-        attach.setId("attach-detach");
+        attachDetach.setId("attach-detach");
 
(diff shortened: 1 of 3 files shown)


pekam:
Actually, this is not the best solution. This most likely fixes #235 only for the first attach.

Because of the way Flow works and creates new elements on attach, calling `initConnector` on attach event would still be the most logical solution. We just need to handle that case when the component is both attached and detached on the same roundtrip, firing a server-side attach event but not actually creating a client-side element.

For this reason, adding the simple falsy-check in comboboxConnector.js would be a fine solution IMO.

pekam:
Closing in favor of #272

PR #286, opened 2019-08-14
ComboBox demo improvement

shadikhani:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/286)

### DIFF
--- a/vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
+++ b/vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
@@ -184,9 +184,8 @@ public class ComboBoxView extends DemoView {
          * Allow users to enter a value which doesn't exist in the data set, and
          * set it as the value of the ComboBox.
          */
-        comboBox.addCustomValueSetListener(event -> {
-            comboBox.setValue(event.getDetail());
-        });
+        comboBox.addCustomValueSetListener(event -> 
+            comboBox.setValue(event.getDetail()));
 
         comboBox.addValueChangeListener(event -> {
             if (event.getValue() == null) {
@@ -265,7 +264,8 @@ public class ComboBoxView extends DemoView {
 
     private void customOptionsDemo() {
         // begin-source-example
-        // source-example-heading: Customizing drop down items with ComponentRenderer
+        // source-example-heading: Customizing drop down items with
+        // ComponentRenderer
         ComboBox<Information> comboBox = new ComboBox<>();
         comboBox.setLabel("User");
         comboBox.setItems(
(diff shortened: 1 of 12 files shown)


vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -41,386 +51,314 @@
...
-         */
-        List<String> names = getNames(500);
-        comboBox.setItems(names);
+        comboBox.setItemLabelGenerator(Information::getText);
         // end-source-example
 
-        comboBox.setId("lazy-loading-box");
-        addCard("Lazy Loading", "Lazy loading between client and server",
-                comboBox);
+        addCard("Presentation",
```
tomivirkki: For some reason the source code of this demo isn't visible on the demo page

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -41,386 +51,314 @@
...
+        comboBox.setId("template-selection-box");
+        addCard("Presentation",
+                "Customizing drop down items with TemplateRenderer", comboBox);
+    }
 
+    private void styling() {
+
+        Div firstDiv = new Div();
+        firstDiv.setText(
+                "To read about styling you can read the related tutorial in");
```
tomivirkki: "To read about X you can read Y in" doesn't sound right.

tomivirkki: Also the text inside the info box doesn't wrap correctly: https://www.dropbox.com/s/v8uinmwliimd870/Screenshot%202019-08-22%2010.08.56.png?dl=0

...should probably look more like https://www.dropbox.com/s/24kwlk8msatcyc9/Screenshot%202019-08-22%2010.07.01.png?dl=0

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
Haprog: This seems to be because `// source-example-heading:` above is wrapped to two lines but only one line is parsed and it needs to match the card heading given for `addCard()`.

Haprog: This is now fixed.

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
Haprog: I pushed an update with the extra "in" word removed, capitalized "HTML" and fixed the markup so it wraps nicely.

The wording in both paragraphs here is still a bit weird and redundant imo but I think now it's kind of "ok". I would maybe prefer to just keep the links but remove the text in front of them since they don't add much value. Or write something completely different instead.

PR #291, opened 2019-09-23
Fix data request range

### LINKED ISSUE #288: Combobox fetches large chunks of data when scrolling very fast
A customer has a combobox backed by a large dataset - if he scrolls by dragging down the index indicator on the right the fetched data sets quickly grow in size slowing down the application.

In a way this is what I would expect since we are scrolling very, very fast. Now, obviously the backend takes a bit of time to fetch all that, stalling the application. However, the increase in the number of fetched items is probably too steep. A general cap on the number of items might be useful here. This also applies to every other component that fetches from a backend.

To give some perspective: I modified the Bakery application starter (Spring Boot) to create 300000 products and then scrolled through the list. The page sizes observed were 50, 50, 50, 2200, 49800, 163000.

tomivirkki:
Fixes #288

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/291)

### DIFF
--- a/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -83,20 +83,21 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
 
         scrollToItem(stringBox, 100);
 
-        waitUntil(e -> getOverlayContents().contains("Item 100"));
+        waitUntil(e -> getOverlayContents().contains("Item 112"));
         assertLoadedItemsCount(
                 "There should be 150 items after loading three pages", 150,
                 stringBox);
-        assertRendered("Item 115");
+        assertRendered("Item 112");
     }
 
     @Test
     public void openPopup_scrollToEnd_onlyFirstAndLastPagesLoaded() {
         stringBox.openPopup();
         scrollToItem(stringBox, 1000);
+        waitUntil(e -> getOverlayContents().contains("Item 999"));
         assertLoadedItemsCount(
-                "Expected the first and the last pages to be loaded (100 items).",
-                100, stringBox);
+                "Expected the first and the last pages to be loaded (50 items).",
+                50, stringBox);
         assertRendered("Item 999");
     }
 
@@ -107,8 +108,8 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         scrollToItem(stringBox, 920);
 
         assertLoadedItemsCount(
-                "Expected the first and the two last pages to be loaded (150 items).",
-                150, stringBox);
+                "Expected the two last pages to be loaded (100 items).",
+                100, stringBox);
         assertRendered("Item 920");
     }
 
@@ -416,9 +417,10 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         clickButton("component-renderer");
         beanBox.openPopup();
         scrollToItem(beanBox, 300);
-        waitUntil(e -> getOverlayContents().contains("<h4>Person 300</h4>"));
+        
+        waitUntil(e -> getOverlayContents().contains("<flow-component-renderer appid=\"ROOT\"><h4>Person 300</h4></flow-component-renderer>"));
         scrollToItem(beanBox, 0);
-        waitUntil(e -> getOverlayContents().contains("<h4>Person 0</h4>"));
+        waitUntil(e -> getOverlayContents().contains("<flow-component-renderer appid=\"ROOT\"><h4>Person 0</h4></flow-component-renderer>"));
 
         assertComponentRendered("<h4>Person 0</h4>");
         assertComponentRendered("<h4>Person 4</h4>");
--- a/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -71,7 +71,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
     }
 
     @Test
-    public void scrollOverlay_morePagesLoaded() {
+    public void scrollOverlay_morePagesLoaded_overflowingPagesDiscarded() {
         stringBox.openPopup();
         scrollToItem(stringBox, 50);
 
@@ -86,6 +86,22 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
                 "There should be 150 items after loading three pages", 150,
                 stringBox);
         assertRendered("Item 102");
+
+        // The first page should get discarded (active range has default limit of 150)
+        scrollToItem(stringBox, 150);
+
+        assertLoadedItemsCount(
+                "There should be 150 items after loading four pages", 150,
+                stringBox);
+        assertRendered("Item 152");
+        
+        // The last page should get discarded (active range has default limit of 150)
+        scrollToItem(stringBox, 0);
+
+        assertLoadedItemsCount(
+                "There should be 150 items after loading four pages", 150,
+                stringBox);
+        assertRendered("Item 2");
     }
 
     @Test
--- a/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -87,19 +87,25 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
                 stringBox);
         assertRendered("Item 102");
 
-        // The first page should get discarded (active range has default limit of 150)
-        scrollToItem(stringBox, 150);
-
-        assertLoadedItemsCount(
-                "There should be 150 items after loading four pages", 150,
-                stringBox);
-        assertRendered("Item 152");
+        // The first pages should get discarded (active range has default limit of 500)
+        for (int i = 150; i <= 600; i += 50) {
+        	scrollToItem(stringBox, i);
+        	waitUntilTextInContent("Item " + i);
+        }
         
-        // The last page should get discarded (active range has default limit of 150)
-        scrollToItem(stringBox, 0);
+        assertLoadedItemsCount(
+                "There should be 500 items after loading multiple pages", 500,
+                stringBox);
+        assertRendered("Item 602");
+        
+        // The last pages should get discarded (active range has default limit of 500)
+        for (int i = 600; i >= 0; i -= 50) {
+        	scrollToItem(stringBox, i);
+        	waitUntilTextInContent("Item " + i);
+        }
 
         assertLoadedItemsCount(
-                "There should be 150 items after loading four pages", 150,
+                "There should be 500 items after scrolling back to start", 500,
                 stringBox);
         assertRendered("Item 2");
     }
@@ -119,7 +125,9 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
     public void scrollToEnd_scrollUpwards_pagesLoaded() {
         stringBox.openPopup();
         scrollToItem(stringBox, 1000);
+        waitUntilTextInContent("Item 999");
         scrollToItem(stringBox, 920);
+        waitUntilTextInContent("Item 919");
 
         assertLoadedItemsCount(
                 "Expected the two last pages to be loaded (100 items).",
@@ -127,6 +135,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         assertRendered("Item 920");
         
         scrollToItem(stringBox, 870);
+        waitUntilTextInContent("Item 869");
         
         assertLoadedItemsCount(
                 "Expected the three last pages to be loaded (150 items).",
@@ -229,6 +238,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
 
         clickButton("change-pagesize");
         pagesizeBox.openPopup();
+        waitUntilTextInContent("Item");
         assertLoadedItemsCount(
                 "After opening the ComboBox, the first 'pageSize' amount "
                         + "of items should be loaded (with updated pageSize: 100).",
--- a/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -83,11 +83,11 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
 
         scrollToItem(stringBox, 100);
 
-        waitUntil(e -> getOverlayContents().contains("Item 112"));
+        waitUntil(e -> getOverlayContents().contains("Item 100"));
         assertLoadedItemsCount(
                 "There should be 150 items after loading three pages", 150,
                 stringBox);
-        assertRendered("Item 112");
+        assertRendered("Item 102");
     }
 
     @Test
@@ -111,6 +111,16 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
                 "Expected the two last pages to be loaded (100 items).",
                 100, stringBox);
         assertRendered("Item 920");
+        
+        scrollToItem(stringBox, 870);
+        
+        assertLoadedItemsCount(
+                "Expected the three last pages to be loaded (150 items).",
+                150, stringBox);
+        assertRendered("Item 870");
+        assertNotRendered("Item 990");
+
+        
     }
 
     @Test
--- a/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -74,6 +74,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
     public void scrollOverlay_morePagesLoaded_overflowingPagesDiscarded() {
         stringBox.openPopup();
         scrollToItem(stringBox, 50);
+        waitUntilTextInContent("Item 52");
 
         assertLoadedItemsCount(
                 "There should be 100 items after loading two pages", 100,
@@ -81,6 +82,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         assertRendered("Item 52");
 
         scrollToItem(stringBox, 100);
+        waitUntilTextInContent("Item 102");
 
         assertLoadedItemsCount(
                 "There should be 150 items after loading three pages", 150,
@@ -225,6 +227,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
                 180, pagesizeBox);
 
         scrollToItem(pagesizeBox, 200);
+        waitUntilTextInContent("Item 200");
 
         assertLoadedItemsCount("Expected two pages to be loaded.", 360,
                 pagesizeBox);
--- a/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -476,8 +476,13 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         beanBox.openPopup();
         waitUntilTextInContent("Person 0");
 
-        scrollToItem(beanBox, 600);
-        waitUntilTextInContent("Person 600");
+        int scrollIndex = 600;
+        scrollToItem(beanBox, scrollIndex);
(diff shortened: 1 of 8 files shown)


vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -105,9 +108,19 @@ public void scrollToEnd_scrollUpwards_pagesLoaded() {
         scrollToItem(stringBox, 920);
 
         assertLoadedItemsCount(
-                "Expected the first and the two last pages to be loaded (150 items).",
-                150, stringBox);
+                "Expected the two last pages to be loaded (100 items).",
+                100, stringBox);
         assertRendered("Item 920");
+        
+        scrollToItem(stringBox, 870);
```
tomivirkki: Extended the test to also scroll backwards to make sure the requested range increases respectively

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -414,7 +427,10 @@ public void setComponentRenderer_scrollDown_scrollUp_itemsRendered() {
         clickButton("component-renderer");
         beanBox.openPopup();
         scrollToItem(beanBox, 300);
+        
+        waitUntil(e -> getOverlayContents().contains("<flow-component-renderer appid=\"ROOT\"><h4>Person 300</h4></flow-component-renderer>"));
```
tomivirkki: Just making sure it actually scrolls to 300

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -25,12 +25,26 @@ window.Vaadin.Flow.comboBoxConnector = {
...
-    comboBox.dataProvider = function (params, callback) {
+    const clearPageCallbacks = () => {
+      // Flush and empty the existing requests
+      Object.keys(pageCallbacks).forEach(page => pageCallbacks[page]([], comboBox.size));
+      pageCallbacks = {};
+
+      // Empty the comboBox's internal cache without invoking observers by filling
+      // the filteredItems array with placeholders (comboBox will request for data when it
+      // encounters a placeholder)
+      const placeHolder = new Vaadin.ComboBoxPlaceholder();
```
tomivirkki: It's more memory efficient to reuse the same placeholder instance

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -56,36 +70,44 @@ window.Vaadin.Flow.comboBoxConnector = {
 
       const filterChanged = params.filter !== lastFilter;
       if (filterChanged) {
-        pageCallbacks = {};
         cache = {};
         lastFilter = params.filter;
+        this._debouncer = Debouncer.debounce(
```
tomivirkki: Moved this up here (with a return statement). It's nicer to invoke `setRequestedRange` from just one place.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -56,36 +70,44 @@ window.Vaadin.Flow.comboBoxConnector = {
...
-                // DataCommunicator thinks it doesn't need to send data
-                comboBox.$server.resetDataCommunicator();
-              }
-            });
+        pageCallbacks[params.page] = callback
+        const activePages = Object.keys(pageCallbacks).map(page => parseInt(page));
+        const rangeMin = Math.min(...activePages);
+        const rangeMax = Math.max(...activePages);
+
+        if (rangeMax - rangeMin + 1 !== activePages.length) {
```
tomivirkki: Let's say the currently active range is from x to y. Only if `params.page` is x - 1 or  y + 1, the range gets extended by the new page. Otherwise the cache is cleared and range gets reset.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -168,10 +190,7 @@ window.Vaadin.Flow.comboBoxConnector = {
         let page = outstandingRequests[i];
 
         if (cache[page]) {
-          let callback = pageCallbacks[page];
-          delete pageCallbacks[page];
```
tomivirkki: Should no longer delete values from the `pageCallbacks` object as it represents the currently active range.

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -85,16 +85,17 @@ public void scrollOverlay_morePagesLoaded() {
...
 
     @Test
     public void openPopup_scrollToEnd_onlyFirstAndLastPagesLoaded() {
         stringBox.openPopup();
         scrollToItem(stringBox, 1000);
+        waitUntil(e -> getOverlayContents().contains("Item 999"));
         assertLoadedItemsCount(
-                "Expected the first and the last pages to be loaded (100 items).",
-                100, stringBox);
+                "Expected the last page to be loaded (50 items).",
```
tomivirkki: The first page is discarded with the new version so there's only 50 items (the last page)

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -414,7 +425,10 @@ public void setComponentRenderer_scrollDown_scrollUp_itemsRendered() {
         clickButton("component-renderer");
         beanBox.openPopup();
         scrollToItem(beanBox, 300);
+        
+        waitUntil(e -> getOverlayContents().contains("<flow-component-renderer appid=\"ROOT\"><h4>Person 300</h4></flow-component-renderer>"));
```
pekam: It would be best to avoid relying on `<flow-component-renderer>` since it's an implementation detail, e.g:
```java
getOverlayContents().stream().anyMatch(s -> s.contains("<h4>Person 300</h4>"))
```

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -85,16 +85,17 @@ public void scrollOverlay_morePagesLoaded() {
         assertLoadedItemsCount(
                 "There should be 150 items after loading three pages", 150,
                 stringBox);
-        assertRendered("Item 115");
+        assertRendered("Item 102");
     }
 
     @Test
     public void openPopup_scrollToEnd_onlyFirstAndLastPagesLoaded() {
```
pekam: The test name should be updated (remove first page).

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
tomivirkki: Done

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
tomivirkki: Done

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -168,10 +190,7 @@ window.Vaadin.Flow.comboBoxConnector = {
         let page = outstandingRequests[i];
```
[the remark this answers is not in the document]
tomivirkki: Done.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -25,12 +25,57 @@ window.Vaadin.Flow.comboBoxConnector = {
...
+            if ((Date.now() - lastRan) >= limit) {
+              func.apply(context, args)
+              lastRan = Date.now()
+            }
+          }, limit - (Date.now() - lastRan))
+        }
+      }
+    };
+    let setRequestedRange;
+    const rangeRequestThrottle = throttle(() => setRequestedRange && setRequestedRange(), 50);
```
pekam: The optimal behavior would probably be debouncing after all, but so that if there's no active timeout, it would make the first call immediately and start the timeout.

Also, the timeout could be bigger. Requests are delayed only when scrolling really fast, and having e.g. 200ms would avoid a lot more requests, without adding too much (IMO) delay to the fast-scrolling case.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -56,36 +79,59 @@ window.Vaadin.Flow.comboBoxConnector = {
...
-                // Fixes the case when the filter changes 
-                // from '' to something else and back to '' 
-                // within debounce timeout, and the
-                // DataCommunicator thinks it doesn't need to send data
-                comboBox.$server.resetDataCommunicator();
-              }
-            });
-        } else {
-          comboBox.$server.setRequestedRange(0, upperLimit, params.filter);
+            timeOut.after(50),
```
pekam: IMO this could be increased to 100 or 200ms. It affects only fast scrolling and I think the benefit of loading less data overcomes the downside of added lag.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -56,36 +79,59 @@ window.Vaadin.Flow.comboBoxConnector = {
...
+          // Wasn't a sequential page index, clear the cache so combo-box will request for new pages
+          clearPageCallbacks();
+        } else {
+          // The requested page was sequential, extend the requested range
+          const startIndex = params.pageSize * rangeMin;
+          const endIndex = params.pageSize * (rangeMax + 1);
+          const count = endIndex - startIndex;
+          setRequestedRange = () => comboBox.$server.setRequestedRange(startIndex, count, params.filter);
+          if (!this._debouncer || !this._debouncer.isActive()) {
+            setRequestedRange();
```
pekam: This immeadiate initial request has a problem. Even if the combo-box should initially load e.g. 2 pages, it loads only the first one immediately and debounces the next request. The same can happen when you quickly scroll to a completely new scroll-position.

This could be fixed by wrapping this part in `setTimeout`, so that the immediately following requests would be included in the first batch. But for some reason, there's a noticeable delay between the `comboBox.dataProvider` calls, which means that the timeout should be several hundreds of milliseconds, which ruins the purpose.

This is not a huge issue, especially in most cases where a single page covers the viewport. We could create a further improvement ticket for this.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -25,12 +25,35 @@ window.Vaadin.Flow.comboBoxConnector = {
 
     comboBox.$connector = {};
 
+    // holds pageIndex -> callback pairs of subsequent indexes (current active range)
     let pageCallbacks = {};
     let cache = {};
     let lastFilter = '';
+    const placeHolder = new Vaadin.ComboBoxPlaceholder();
+    const MAX_RANGE_COUNT = Math.max(comboBox.pageSize * 2, 150); // Max item count in active range
```
pekam: This could also be increased, e.g. to 300 or 500 to avoid re-loading data too much (like you suggested yourself IIRC).

Actually I'm not even sure anymore if this is needed with the current implementation. What was the reason again to dump old pages also when not skipping pages? 😅 Well, it works the same way as Grid, so probably fine this way.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -25,12 +25,35 @@ window.Vaadin.Flow.comboBoxConnector = {
...
+        pageCallbacks[page]([], comboBox.size);
+        delete pageCallbacks[page];
+
+        // Empty the comboBox's internal cache without invoking observers by filling
+        // the filteredItems array with placeholders (comboBox will request for data when it
+        // encounters a placeholder)
+        const pageStart = parseInt(page) * comboBox.pageSize;
+        const pageEnd = pageStart + comboBox.pageSize;
+        for (let i = 0; i < comboBox.filteredItems.length; i++) {
+          if (i >= pageStart && i < pageEnd) {
```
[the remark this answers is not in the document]
tomivirkki: Good catch, yeah the check is still needed as the last page might not be complete. Done.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
tomivirkki: Done.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
tomivirkki: > What was the reason again to dump old pages also when not skipping pages?

If you use one of those super mice you have, you can quickly scroll down and buffer a request of thousands of items before it's sent. Increased to 500.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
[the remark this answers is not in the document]
pekam: Seems that you've used tabs instead of spaces for indentation.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
pekam: Right 👍

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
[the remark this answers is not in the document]
tomivirkki: Fixed.

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -71,43 +71,81 @@ public void openPopup_firstPageLoaded() {
...
+        scrollToItem(stringBox, 870);
+        waitUntilTextInContent("Item 869");
+        
+        assertLoadedItemsCount(
+                "Expected the three last pages to be loaded (150 items).",
+                150, stringBox);
+        assertRendered("Item 870");
+        assertNotRendered("Item 990");
+
+
```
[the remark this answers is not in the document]
tomivirkki: Done. Somehow my IDE had lost the formatting rules.

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -414,7 +454,10 @@ public void setComponentRenderer_scrollDown_scrollUp_itemsRendered() {
         clickButton("component-renderer");
         beanBox.openPopup();
         scrollToItem(beanBox, 300);
+        
+        waitUntil(e -> getOverlayContents().stream().anyMatch(s -> s.contains("<h4>Person 300</h4>")));
```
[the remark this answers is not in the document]
tomivirkki: Done.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -25,12 +25,35 @@ window.Vaadin.Flow.comboBoxConnector = {
...
+    // holds pageIndex -> callback pairs of subsequent indexes (current active range)
     let pageCallbacks = {};
     let cache = {};
     let lastFilter = '';
+    const placeHolder = new Vaadin.ComboBoxPlaceholder();
+    const MAX_RANGE_COUNT = Math.max(comboBox.pageSize * 2, 500); // Max item count in active range
+
+    let setRequestedRange;
+
+    const clearPageCallbacks = (pages = Object.keys(pageCallbacks)) => {
```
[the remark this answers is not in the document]
tomivirkki: Done. Added a call to `refreshValue()` to `setRequestedRange` and a test case to cover this case.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
tomivirkki: Adding a timeout longer than a few milliseconds defeats the purpose of the initial call as you said so let's keep this as it is

pekam: That's why it should be fixed by removing the delay between `dataProvider` calls. So that if combo-box needs two pages, it would immediately call `dataProvider` two times.

I will at least create a ticket so we don't lose the reference to this discussion.

pekam: Nevermind. I think it was a noticable issue only when I was testing with really long debounce timeouts. With 200ms in the current implementation it's fine.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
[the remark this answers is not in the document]
tomivirkki: Looks like `refreshValue()` will not only update the component's value but also override any filters a user might have typed to the field. Will need to find a way to avoid this to allow filtering while a value is selected.

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -413,14 +455,36 @@ public void filterAndSelectItemNotOnFirstPage_setCurrentValue_valueCorrect() {
...
+    public void scrollDown_scrollUp_selectionRetained() {
+        beanBox.sendKeys("Person 0");
+        waitUntilTextInContent("Person 0");
+        beanBox.sendKeys(Keys.ENTER);
+
+        beanBox.openPopup();
+        waitUntilTextInContent("Person 0");
+
+        scrollToItem(beanBox, 600);
+        waitUntilTextInContent("Person 600");
```
pekam: This might turn into a false-positive in the future if we decide to increase the amount of items to keep in client-side from 500.

Would be more robust if we check that `getLoadedItems` (which basically returns `filteredItems`) doesn't include the first item anymore.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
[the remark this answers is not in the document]
tomivirkki: Ok changed the logic to maintain the selectedItem key (and value) in the connector instead. Added a test to cover the mentioned regression.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -56,36 +79,59 @@ window.Vaadin.Flow.comboBoxConnector = {
...
+          () => {
+            clearPageCallbacks();
+            comboBox.clearCache();
+            if (params.filter === '') {
+              // Fixes the case when the filter changes
+              // from '' to something else and back to ''
+              // within debounce timeout, and the
+              // DataCommunicator thinks it doesn't need to send data
+              comboBox.$server.resetDataCommunicator();
+            }
```
pekam: Not sure if it's regression from these changes or not, but setting a filter, and then changing it back and forth quickly to the old filter leaves the combo-box in a loading state.
1. type `1` and wait for the data to load
2. type quickly any key + backspace so the filter is again `1`
3. notice that loading indicator keeps spinning and data is not visible

The server gets these requests:
0-50 filter: "1"
0-50 filter: "1"
Probably it thinks that no data needs to be sent because nothing has changes since the last call.

(not necessarily related to the highlighted lines, they seem to fix the same issue but only for empty filter)

pekam: Not a regression, so doesn't block this PR. Probably the same issue as in #296, where I've added a comment.

PR #302, opened 2019-10-16
Fix for filtering problem using slow DataProvider - 296

### LINKED ISSUE #296: Filtering problem using slow DataProvider
### Vaadin Framework version
Tested on Vaadin 14.0.6 and 14.0.7
### Browser version
Chrome Version 77.0.3865.90
### Web container name and version
Jetty 9.4.15.v20190215
### Description of the bug
When using a DataProvider that takes half a second to load the data from backend, it seems to be a race condition when losing and getting focus again, after selecting an item.
### Minimal reproducible example
[See attached project](https://github.com/vaadin/vaadin-combo-box-flow/files/3672807/combobox-problem.zip)
Animated gif showing the problem:
ComboProblem
It can be reproduced by doing this with only the keyboard:
1) Focus on the combobox
2) Press "a", wait until the item is shown and then select it by "down" and "enter"
3) Press "tab" to loose focus
4) Press "shift"+"tab" to gain focus again
5) Press "a" again
### Expected behavior
The combobox should show the filtered item
### Actual behavior
The loading indicator remains visible and the filtered item does not appear

tulioag:
Fixes #296 

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box-flow/302)

### DIFF
--- a/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
+++ b/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
@@ -47,7 +47,7 @@ window.Vaadin.Flow.comboBoxConnector = {
           lastFilterSentToServer = filter;
           if(dataCommunicatorResetNeeded) {
             comboBox.$server.resetDataCommunicator();
-            dataCommunicatorResetNeeded = true;
+            dataCommunicatorResetNeeded = false;
           }
         };
 
--- a/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
+++ b/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
@@ -33,6 +33,7 @@ window.Vaadin.Flow.comboBoxConnector = {
     const MAX_RANGE_COUNT = Math.max(comboBox.pageSize * 2, 500); // Max item count in active range
 
     let setRequestedRange;
+    let needsDataCommunicatorReset;
 
     const clearPageCallbacks = (pages = Object.keys(pageCallbacks)) => {
       // Flush and empty the existing requests
@@ -81,18 +82,20 @@ window.Vaadin.Flow.comboBoxConnector = {
       if (filterChanged) {
         cache = {};
         lastFilter = params.filter;
+        needsDataCommunicatorReset = needsDataCommunicatorReset || params.filter === '';
         this._debouncer = Debouncer.debounce(
           this._debouncer,
           timeOut.after(500),
           () => {
             clearPageCallbacks();
             comboBox.clearCache();
-            if (params.filter === '') {
+            if (needsDataCommunicatorReset) {
               // Fixes the case when the filter changes
               // from '' to something else and back to ''
               // within debounce timeout, and the
               // DataCommunicator thinks it doesn't need to send data
               comboBox.$server.resetDataCommunicator();
+              needsDataCommunicatorReset = false;
             }
           });
         return;
--- a/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
+++ b/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
@@ -25,14 +25,35 @@ window.Vaadin.Flow.comboBoxConnector = {
     comboBox.$connector = {};
 
     // holds pageIndex -> callback pairs of subsequent indexes (current active range)
-    let pageCallbacks = {};
+    const pageCallbacks = {};
     let cache = {};
     let lastFilter = '';
     const placeHolder = new Vaadin.ComboBoxPlaceholder();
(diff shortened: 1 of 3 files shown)


vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -81,18 +82,20 @@ window.Vaadin.Flow.comboBoxConnector = {
       if (filterChanged) {
         cache = {};
         lastFilter = params.filter;
+        needsDataCommunicatorReset = needsDataCommunicatorReset || params.filter === '';
```
pekam: This seems to fix the described issue, but not the related problem which I described [in the comment](https://github.com/vaadin/vaadin-combo-box-flow/issues/296#issuecomment-537833465). That would be fine of course, we could create another ticket for that.

But I have a problem with this solution. If I've understood correctly, it works only because of a bug in the web component:
When you move focus out and back into the combo-box, it is not opened and thus `dataProvider` is not called. Now when you type "a" into the box, you'd expect a `dataProvider` call with `filter: "a"`. But before that, there's a `dataProvider` call with `filter: ""`. If I've understood correctly, this empty-filter call, which shouldn't be there, is the reason why this solution fixes the issue.

I think the root problem is that when the server gets the same arguments as last time (in `setRequestedRange`), it won't send anything, because it thinks it answered to that call already. But client has dumped that data.

Maybe a better way to decide `needsDataCommunicatorReset` would be to:
- save latest filter which is actually sent to the server, when calling `setRequestedRange`
- compare the current filter to that latest committed filter in the debouncer
- if we are going to make a request with the same filter as the last one actually sent to server, data communicator needs to be reset

WDYT? Did I misunderstand something in your solution?

pekam: I tested the latest approach a bit more and noticed that now `resetDataCommunicator` is called each time when the filter changes, not only when it changes back to `lastFilterSentToServer`.

It works fine, as it is ok to reset everything when the filter changes. But that is a bit ugly and maybe bad for server-side performance to call that method too often (not sure though).

I think the problem is that this part:
```js
clearPageCallbacks();
comboBox.clearCache();
```
triggers the `dataProvider` again, which saves the latest filter, and after that the filter-check is done, so it compares the current filter to the same one which was already sent to server.

Anyway, moving that part after the if-block seems to fix the issue. So it first compares the filters, and then calls `clearCache` to trigger new `dataProvider` call.

tulioag: I refactored a bit to solve this issue. In the end, there seems to be a few corner cases where the filter is not properly applied, but, at least in the ones I was able to reproduce, the problem was that `combox.filter` has an empty value instead of the value typed in the search box. To me that sounds like an issue in the component itself.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -22,17 +22,38 @@ window.Vaadin.Flow.comboBoxConnector = {
...
+        const getLastFilterSentToServer = () => lastFilterSentToServer;
+        const requestData = (startIndex, endIndex, params) => {
+          const count = endIndex - startIndex;
+          const filter = params.filter;
+
+          comboBox.$server.setRequestedRange(startIndex, count, filter);
+          lastFilterSentToServer = filter;
+          if(dataCommunicatorResetNeeded) {
+            comboBox.$server.resetDataCommunicator();
+            dataCommunicatorResetNeeded = true;
```
pekam: This should probably be `false`?

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
pekam: I reported this bug to web component: https://github.com/vaadin/vaadin-combo-box/issues/848
Not sure if there are other problematic cases not covered by this ticket.

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
tulioag: Done

vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
tulioag: In this branch:
1. Go to http://localhost:9998/filter-debounce
1. Select the value "aaa"
1. Type backspace 3 times then "b" quickly
1. Wait a bit
1. Type backspace and "a"
1. Wait a bit

Result: The filter will be empty, although the value in the input box is "a".

pekam: I created a ticket and found the cause of the problem:
https://github.com/vaadin/vaadin-combo-box-flow/issues/304

It's from the earlier rework of the connector logic.

PR #348, opened 2020-04-30
Update demos order

alvarezguille:
Keep custom value demos together


alvarezguille:
Closed in favor of #352

PR #351, opened 2020-05-11
Add add-method call to make code complete

anasmi:
For users to be able to use code correctly by copy-pasting
Depends on https://github.com/vaadin/flow/pull/8199/files

Also fixes https://github.com/vaadin/vaadin-combo-box-flow/issues/353


anasmi:
@alvarezguille done :)

Issue #355, opened 2020-05-12
Take DataView concept into use


caalador:
Based on vaadin/flow#8303 and the common implementation parts introduced there, take ListDataView into use with ComboBox. Make a PR to feature/dataprovider branch, use the 3.2.dataprovider-SNAPSHOT version of Flow.

The demo for data view should showcase:

 - migrating all existing demos to use data view-based API instead
 - setting items and creating next/previous controls
 - filtering and sorting

mstahv:
Removed sorting from the list as there is no sorting in ComboBox (but what developer does).

mshabarov:
As it figured out during the initial attempt to implement `HasXXXDataView` to `ComboBox`, there are two obstacles to do that smoothly:

- `HasDataView` contains `setItems(InMemoryDataProvider<T> dataProvider)` which does not match with an existing `setDataProvider` methods, because `InMemoryDataProvider` has a `SerializablePredicate<T>` as a filter type and `ComboBox` has a `String`
- `HasLazyDataView` has a `Void` as a filter type, so it cannot be applied to `ComboBox` as is.

The first issue was resolved just by skipping the `HasDataView` for `ComboBox`, and the second required to create a new interface `HasFilterableLazyDataView` with an additional generic type representing the filter type:

```
public interface HasFilterableLazyDataView<T, F, V extends FilterableLazyDataView<T, F>>
        extends Serializable {

    default V setItemsWithFilter(
            CallbackDataProvider.FetchCallback<T, F> fetchCallback) {
        return setItemsWithConvertedFilter(fetchCallback,
                SerializableFunction.identity());
    }

    default V setItemsWithFilter(
            CallbackDataProvider.FetchCallback<T, F> fetchCallback,
            CallbackDataProvider.CountCallback<T, F> countCallback) {
        return setItemsWithFilter(DataProvider
                .fromFilteringCallbacks(fetchCallback, countCallback));
    }

    default <Q> V setItemsWithConvertedFilter(
            CallbackDataProvider.FetchCallback<T, Q> fetchCallback,
            SerializableFunction<F, Q> filterConverter) {
        V filterableLazyDataView = setItemsWithConvertedFilter(fetchCallback,
                query -> {
                    throw new IllegalStateException(
                            "Trying to use exact size with a lazy loading component"
                                    + " without either providing a count callback for the"
                                    + " component to fetch the count of the items or a data"
                                    + " provider that implements the size query. Provide the "
[...]

PR #380, opened 2020-07-31
Take DataView concept into use

mshabarov:
Part of #355


vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -226,10 +227,7 @@ private void storingCustomValues() {
         // begin-source-example
         // source-example-heading: Storing custom values
         ComboBox<Project> comboBox = new ComboBox<>("Project");
-        DataProvider<Project, String> dataProvider = DataProvider
-                .fromFilteringCallbacks(this::fetchProjects,
-                        this::countProjects);
-        comboBox.setDataProvider(dataProvider);
+        comboBox.setItems(this::fetchProjects, this::countProjects);
```
pleku, line 230: Just as a note - I think the CB demos should be similar to grid - show the usage without the count query and then show how to get to explicit count. And maybe even the estimate API.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -385,30 +394,6 @@ public void setRenderer(Renderer<T> renderer) {
...
-     * page size with {@link #setPageSize(int)}.
-     * <p>
-     * Setting the items creates a new DataProvider, which in turn resets the
-     * combo box's value to {@code null}. If you want to add and remove items to
-     * the current item set without resetting the value, you should update the
-     * previously set item collection and call
-     * {@code getDataProvider().refreshAll()}.
-     */
-    @Override
-    public void setItems(Collection<T> items) {
```
pleku, line 408: I think you should retain overriding this method by defer to super implementation because the javadocs explain how this method works with CB, while the super method does not

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -423,17 +408,25 @@ public void setItems(Collection<T> items) {
      * the current item set without resetting the value, you should update the
      * previously set item collection and call
      * {@code getDataProvider().refreshAll()}.
+     * <p>
+     * The returned data view object can be used for further access to ComboBox
+     * items, or later on fetched with {@link #getListDataView()}. For using
+     * lazy data load, use one of the {@code setItemsWithFilter} instead.
```
pleku, line 414: wat ? wrong method name ?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -450,16 +443,111 @@ public void setItems(ItemFilter<T> itemFilter, Collection<T> items) {
...
```
pleku, line 449: wat ? wrong method name ?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -450,16 +443,111 @@ public void setItems(ItemFilter<T> itemFilter, Collection<T> items) {
...
             @SuppressWarnings("unchecked") T... items) {
-        setItems(itemFilter, Arrays.asList(items));
+        return setItems(itemFilter, Arrays.asList(items));
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @deprecated Because the stream is collected to a list anyway, use
+     *             {@link HasListDataView#setItems(Collection)} or
```
pleku, line 469: Link to be fixed after this method is added back

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -450,16 +443,111 @@ public void setItems(ItemFilter<T> itemFilter, Collection<T> items) {
...
+
+    @Override
+    public ComboBoxDataView<T> setItems(DataProvider<T, String> dataProvider) {
+        setDataProvider(dataProvider);
+        return getGenericDataView();
+    }
+
+    /**
+     * Gets the generic data view for the ComboBox. This data view should
+     * only be used when {@link #getListDataView()} or
```
pleku, line 487: "only used" is not correct since this can be used when you don't want to distinct between which type it is

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -450,16 +443,111 @@ public void setItems(ItemFilter<T> itemFilter, Collection<T> items) {
...
+        return getLazyDataView();
+    }
+
+    /**
+     * Gets the lazy data view for the ComboBox. This data view should only be
+     * used when the items are provided lazily from the backend with:
+     * <ul>
+     * <li>{@link #setItems(CallbackDataProvider.FetchCallback)}</li>
+     * <li>{@link #setItems(CallbackDataProvider.FetchCallback, CallbackDataProvider.CountCallback)}</li>
+     * <li>{@link #setItems(BackEndDataProvider)}</li>
```
pleku, line 512: Missing the API which accepts callback + filter converter ?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -476,12 +564,43 @@ public void setItems(ItemFilter<T> itemFilter,
...
+     *             access to either
+     *             {@link ComboBoxListDataView} or {@link ComboBoxLazyDataView}
      */
-    @Override
+    @Deprecated
     public void setDataProvider(DataProvider<T, String> dataProvider) {
         setDataProvider(dataProvider, SerializableFunction.identity());
     }
 
+    public <C> ComboBoxLazyDataView<T> setItems(CallbackDataProvider.FetchCallback<T, C> fetchCallback,
```
pleku, line 577: Missing javadocs

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -476,12 +564,43 @@ public void setItems(ItemFilter<T> itemFilter,
...
+                            + "comboBox.getLazyDataView().withDefinedSize(CallbackDataProvider.CountCallback);"
+                            + "%nor switch to undefined size with%n"
+                            + "comboBox.getLazyDataView().withUndefinedSize();");
+        }), filterConverter);
+        ComboBoxLazyDataView<T> lazyDataView = getLazyDataView();
+        lazyDataView.setItemCountUnknown();
+        return lazyDataView;
+    }
+
+    public <C> ComboBoxLazyDataView<T> setItems(CallbackDataProvider.FetchCallback<T, C> fetchCallback,
```
pleku, line 596: Missing javadocs

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -621,6 +753,36 @@ public void setDataProvider(FetchItemsCallback<T> fetchItems,
...
+     *            a callback for fetching items
+     * @param sizeCallback
+     *            a callback for getting the count of items
+     * @return ComboBoxLazyDataView instance for further configuration
+     */
+    // TODO: this methods clashes with
+    // setItems(CallbackDataProvider.FetchCallback<T, F> fetchCallback,
+    //            CallbackDataProvider.CountCallback<T, F> countCallback)
+    public ComboBoxLazyDataView<T> setItems(FetchItemsCallback<T> fetchItems,
+            SerializableFunction<String, Integer> sizeCallback) {
```
pleku, line 781: Probably we'll just have to break this one for unity with other components API.
Or another option to reverse the ordering of the callbacks, but that would still be breaking, so maybe just remove it.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -639,16 +801,27 @@ public void setDataProvider(FetchItemsCallback<T> fetchItems,
...
+                                ListDataProvider<T> listDataProvider) {
         Objects.requireNonNull(listDataProvider,
                 "List data provider cannot be null");
 
         setDataProvider(listDataProvider,
                 filterText -> item -> itemFilter.test(item, filterText));
     }
 
+
+    public ComboBoxListDataView<T> setItems(ItemFilter<T> itemFilter,
```
pleku, line 819: Missing javadocs

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxListDataViewTest.java
```diff
@@ -0,0 +1,146 @@
...
+        public int hashCode() {
+            return Objects.hash(id, value);
+        }
+
+        @Override
+        public String toString() {
+            return String.valueOf(value);
+        }
+    }
+}
```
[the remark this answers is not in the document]
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: This method has been removed to get rid of conflicts. Overload `setItems(FetchCallback<T, F> fetchCallback, CountCallback<T, F> countCallback)` can be used instead

mshabarov:
IT tests for lazy loading are failing, need an investigation and fix

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
mshabarov: Yes, all those mentioned demos added.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Ooops...Done.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -335,6 +413,38 @@ private void customFiltering() {
 
     }
 
+    private void filteringAndSortingWithDataView() {
+        // begin-source-example
+        // source-example-heading: Filtering and Sorting with Data View
+        ComboBox<Person> comboBox = new ComboBox<>("Persons");
+        PersonService personService = new PersonService();
```
pleku, line 420: Maybe just store the collection instead of the service to make it obvious that this is in-memory example

pleku (changes requested):
Partly reviewed, demos need changes

pleku (changes requested):

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/AbstractItemCountComboBoxIT.java
```diff
@@ -54,13 +54,12 @@ protected void open(int size) {
...
-        // TODO: Is that happens for ComboBox too?
-        // FIXME when grid reduces size, it does currently some extra fetches
-        // -> not checking the requested items until this is fixed
-        // verifyFetchForUndefinedSizeCallback(fetchIndex,
-        // Range.between(start, end));
+            int start, int end, String expectedItemText) {
+        scrollToItemAndVerifyLabel(comboBoxElement, itemToScroll, expectedItemText);
+        // TODO: give an array of ranges as a parameter for this method,
+        //  because combo box might scroll to the target item several times
+        // verifyFetchForUndefinedSizeCallback(fetchIndex, Range.between(start,
```
pleku, line 61: Missing the ticket # / link ?

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
```diff
@@ -64,43 +65,37 @@ public void undefinedSize_switchesToDefinedSize_sizeChanges() {
...
         // increase backend size and scroll to current end
         setUnknownCountBackendSize(2000);
         // size has been increased again by default size
-        doScroll(1000, 1200, 6, 950, 1100);
-
-        Assert.assertEquals(1001, getItems(comboBoxElement).size());
+        doScroll(1000, 1200, 6, 950, 1100, "Callback Item 1000");
     }
 
     // @Test TODO
```
pleku, line 101: Still TODOs present

pleku (changes requested):
Also previous comments in demo left unhandled

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -84,35 +96,39 @@
...
+     * non-null filter, offset index and limit.
      *
      * @param <T>
      *            item (bean) type in ComboBox
+     * @param <F>
+     *            filter type
      */
     @FunctionalInterface
-    public interface FetchItemsCallback<T> extends Serializable {
+    public interface FetchItemsCallback<T, F> extends Serializable {
```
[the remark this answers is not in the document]
mshabarov: > So existing usage of String for the filter type should compile and get the same behavior, right ?
Yes, it should. From the client code perspective, it should be fine. They can use `void setDataProvider(FetchItemsCallback<T, String> fetchItems, SerializableFunction<String, Integer> sizeCallback)` as usual. From binary compatibility, it should be also fine, because `F` becomes `Object` after erasure, and it can obtain `String`.

> Another issue might be for the component which extends and reuses this
Yes, that is inconvenience. But from the other side. We have those two methods clashing:

- `setItems(CallbackDataProvider.FetchCallback<T, C> fetchCallback, SerializableFunction<String, C> filterConverter)`
- `setItems(CallbackDataProvider.FetchCallback<T, C> fetchCallback, CallbackDataProvider.CountCallback<T, C> countCallback`)

Compiler can't decide which one to use for this, because `can` be interpreted as a `Integer`:

```
        comboBox.setItems(query -> service.fetch(...), query -> 42);
```

Even swapping the arguments won't help. I see no other elegant solution for that so far.

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -254,29 +259,101 @@ private void storingCustomValues() {
...
+        // https://github.com/vaadin/vaadin-combo-box-flow/tree/master/vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/service/PersonService.java
+        // The backend will have 12345 items
+        PersonService service = new PersonService(12345);
+        ComboBox<Person> comboBox = new ComboBox<>();
+
+        ComboBoxLazyDataView<Person> lazyDataView = comboBox.setItems(
+                service::fetch);
+
+        /*
+         * By default the grid will initially adjust the scrollbar to 200 items
```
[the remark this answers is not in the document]
mshabarov: Oh, sorry. Done

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
mshabarov: Well, I'd better prefer explicitly create a variable and bind it:

```
Collection<Person> persons = personService.fetchAll();
```

because this example requires many `Person` item to represent the filtering and sorting, so I don't really want to make it cumbersome in the example code.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -418,22 +431,27 @@ public void setItems(Collection<T> items) {
...
-     * Setting the items creates a new DataProvider, which in turn resets the
-     * combo box's value to {@code null}. If you want to add and remove items to
-     * the current item set without resetting the value, you should update the
-     * previously set item collection and call
-     * {@code getDataProvider().refreshAll()}.
+     * Setting the items resets the combo box's value to {@code null}.
+     * <p>
+     * The returned data view object can be used for further access to combo box
+     * items, or later on fetched with {@link #getListDataView()}. For using
+     * lazy data load, use one of the {@code setItems} overloaded methods whose
```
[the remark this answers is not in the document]
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -418,22 +431,27 @@ public void setItems(Collection<T> items) {
...
+     * take a fetch callback as a parameter instead.
```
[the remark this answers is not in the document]
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -445,21 +463,186 @@ public void setItems(ItemFilter<T> itemFilter, Collection<T> items) {
...
```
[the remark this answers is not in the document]
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -445,21 +463,186 @@ public void setItems(ItemFilter<T> itemFilter, Collection<T> items) {
...
+     * <p>
+     * Use another overloaded method with filter converter
+     * {@link #setItems(InMemoryDataProvider, SerializableFunction)}
+     * 
+     * @throws UnsupportedOperationException
+     *
+     * @see #setItems(InMemoryDataProvider, SerializableFunction)
+     */
+    @Override
+    public ComboBoxDataView<T> setItems(InMemoryDataProvider<T> dataProvider) {
```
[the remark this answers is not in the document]
mshabarov: Done

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
```diff
@@ -0,0 +1,125 @@
...
+        // jump over a page, trigger fetch
+        doScroll(270, 400, 3, 250, 350);
+
+        // trigger another buffer increase but not capping size
+        doScroll(395, 600, 4, 350, 450);
+
+        // scroll to actual end, no more items returned and size is adjusted
+        doScroll(500, 500, 5, 450, 500);
+        Assert.assertEquals(499, getItems(comboBoxElement).size());
+        // TODO Is it relevant to ComboBox?
```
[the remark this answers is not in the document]
mshabarov: Umm, no, I just wanted to check if _TODO #1038 test further after grid_ is also applicable for ComboBox

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
```diff
@@ -0,0 +1,125 @@
...
+        // TODO #1038 test further after grid is note fetching extra stuff when
+        // size has been adjusted to less than what it is
```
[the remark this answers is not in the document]
mshabarov: As it turned out, this TODO is specific to grid, so it was removed in CB.

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
```diff
@@ -0,0 +1,125 @@
...
+
+        doScroll(120, 400, 1, 50, 200);
+
+        doScroll(299, actualSize, 2, 150, actualSize);
+
+        Assert.assertEquals(300, getItems(comboBoxElement).size());
+
+        // change callback backend size limit
+        setUnknownCountBackendSize(DEFAULT_DATA_PROVIDER_SIZE);
+        // TODO is that relevant to ComboBox?
```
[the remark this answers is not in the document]
mshabarov: Removed. I added this comment just to check if CB should have the same test and same behaviour. It should.

vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/dataview/AbstractItemCountComboBoxPage.java
```diff
@@ -0,0 +1,266 @@
...
+                            .toString());
+            Logger.getLogger(getClass().getName()).info(String.format(
+                    "DataProvider Query : limit %s offset %s", limit, offset));
+//            return IntStream.range(0, DEFAULT_DATA_PROVIDER_SIZE)
+//                    .mapToObj(index -> "DataProvider Item " + index)
+//                    .filter(item -> query.getFilter()
+//                            .map(filter -> item.toLowerCase()
+//                                    .contains(filter.toLowerCase()))
+//                            .orElse(true))
+//                    .skip(offset).limit(limit);
```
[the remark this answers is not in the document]
mshabarov: Removed. I was intending to add a client filter + lazy load with unknown size, but not sure if its really needed. I will check whether the defined size lazy loading is checked versus client filter, or not. If yes, I suppose we don't need add anything more.

vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/dataview/AbstractItemCountComboBoxPage.java
```diff
@@ -0,0 +1,266 @@
...
+                .format("Callback Query : limit %s offset %s", limit, offset));
+        return IntStream.range(offset, lastItemToFetch)
+                .mapToObj(index -> "Callback Item " + index);
+//        return IntStream.range(0, DEFAULT_DATA_PROVIDER_SIZE)
+//                .mapToObj(index -> "Callback Item " + index)
+//                .filter(item -> query.getFilter()
+//                        .map(filter -> item.toLowerCase()
+//                                .contains(filter.toLowerCase()))
+//                        .orElse(true))
+//                .skip(offset).limit(lastItemToFetch);
```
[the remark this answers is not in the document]
mshabarov: Done

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -254,29 +259,101 @@ private void storingCustomValues() {
     private void lazyLoading() {
         //@formatter:off
         // begin-source-example
-        // source-example-heading: Lazy loading with callbacks
+        // source-example-heading: Lazy Loading with Callback
+        // PersonService can be found:
+        // https://github.com/vaadin/vaadin-combo-box-flow/tree/master/vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/service/PersonService.java
+
+        ComboBox<Person> comboBox = new ComboBox<>();
+        PersonService service = new PersonService(500);
```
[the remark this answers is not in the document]
mshabarov: Not done yet. Still in progress...

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -658,6 +987,15 @@ public void setDataProvider(ItemFilter<T> itemFilter,
         return dataCommunicator.getDataProvider();
     }
 
+    /**
+     * Returns the data communicator of this ComboBox.
+     *
+     * @return the data communicator, not {@code null}
+     */
+    public DataCommunicator<T> getDataCommunicator() {
```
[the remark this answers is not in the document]
mshabarov: Done

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxListDataViewTest.java
```diff
@@ -0,0 +1,146 @@
...
+        @Override
+        public int hashCode() {
+            return Objects.hash(id, value);
+        }
+
+        @Override
+        public String toString() {
+            return String.valueOf(value);
+        }
+    }
```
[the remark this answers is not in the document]
mshabarov: Done. `getItems_withInMemoryAndTextFilter_filteredItemsObtained` handles this.

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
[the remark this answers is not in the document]
mshabarov: Removed this TODO

mshabarov:
Changes in custom inmemory data provider https://github.com/vaadin/flow/pull/8961

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/AbstractItemCountComboBoxIT.java
mshabarov: Done

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
mshabarov: Needs a clarification what steps to add in these tests, let's discuss it when appropriate.

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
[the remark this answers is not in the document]
mshabarov: I reverted the changes in `PersonService` so as to retain initial API and exclude Vaadin-specific `Query` class from it. But due to a `setItems` clash, I was forced to use a lambda, instead of method reference (`service::fetch`), which probably reduce the readability and code 'sugar'.

mshabarov:
I also excluded these tests which were supposed to be implemented previously:
- _undefinedSize_switchesToEstimateCallback_sizeChanges_ (no estimate callback is present anymore in the LazyDataView API and switching to a new estimation is already being tested within corresponding IT test class)
- _undefinedSize_switchesToEstimateCallbackSizeLessThanCurrent_throws_ (no estimate callback is present anymore in LazyDataView API and there is no functionality for estimation increase which throws in such a circumstances).

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxListDataViewTest.java
```diff
@@ -0,0 +1,213 @@
...
+                dataView.addItemCountChangeListener(event -> {});
+
+        // Trigger ComboBox::refreshAllData()
+        dataView.setFilter(item -> true);
+
+        registration1.remove();
+        Assert.assertFalse(Boolean.parseBoolean(getClientSideFilter()));
+
+        registration2.remove();
+        Assert.assertTrue(Boolean.parseBoolean(getClientSideFilter()));
```
pleku, line 137: I think I commented on this exact same test elsewhere, I don't get how count listener and filtering are related ?

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxListDataViewTest.java
```diff
@@ -0,0 +1,213 @@
...
+    @Test
+    public void getItems_withInMemoryAndTextFilter_filteredItemsObtained() {
+        ComboBox<String> comboBox = getDefaultLocaleComboBox();
+        setComboBoxTextFilter(comboBox, "ba");
+        items = Arrays.asList("foo", "bar", "banana");
+        ComboBoxListDataView<String> dataView = comboBox.setItems(items);
+        dataView.setFilter(item -> item.length() == 3);
+        Stream<String> allItems = dataView.getItems();
+        Assert.assertArrayEquals("Unexpected data set",
+                new String[] { "bar" }, allItems.toArray());
```
pleku, line 68: This is partly a nit, but I see that a lot of times your tests only verify that "if a, then b happens", but not after again that "if a is removed, then c happens" which for instance in this case means that you're not testing changing/removing filters work at all. I'm pretty sure that those work, but as an overall comment, I've seen that is often not tested and very often also the source of bugs that users report, which could have been caught with very little testing effort.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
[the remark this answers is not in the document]
pleku: It has been too long time since the initial review - did we sort this out somehow ?
Why did we choose to have `F` type instead of `String` which the CB always provides and then force the conversion to happen in the callback implementation instead ?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -147,6 +177,10 @@ public void remove() {
 
         private UpdateQueue(int size) {
             enqueue("$connector.updateSize", size);
+
+            // Update the server side property with the latest value of items
+            // count
+            getElement().setProperty("size", size);
```
pleku, line 183: Uhm, why is this needed ? Please answer in code comment

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -207,8 +241,15 @@ public void initialize() {
     // provided.
     private String lastFilter;
 
+    /*
+     * Holds the item count listeners count.
+     * If the value > 0, then the client side filter will be sent to server
+     * forcibly, in order to let the items size updated and item count change
+     * event triggered (if, of course, the size has changed).
```
pleku, line 248: Isn't client side filtering used only with in-memory data ? It seems just very odd to me that the whole filtering mode would change when the only reason why this is necessary is that the server side needs to know the amount of items.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -530,38 +888,32 @@ public void setDataProvider(DataProvider<T, String> dataProvider) {
...
-        // Postpone data communicator provider initialization in order to
-        // trigger item count request and data fetch upon clicking on combobox
-        dataCommunicatorInitializer = () -> {
-            dataCommunicatorInitializer = null;
-            if (lazyOpenRegistration != null) {
-                lazyOpenRegistration.remove();
-                lazyOpenRegistration = null;
+        if (lazyOpenRegistration != null) {
+            lazyOpenRegistration.remove();
+            lazyOpenRegistration = null;
```
pleku, line 893: Why is this cleared and added again at the end of this method ?
Arguably should have not been added (again) at least when new data provider is set but not sure why it should be cleared if it not null

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -618,8 +967,14 @@ protected void onDetach(DetachEvent detachEvent) {
...
-        setClientSideFilter(!forceServerSideFiltering && getDataProvider()
-                .size(new Query<>()) <= getPageSizeDouble());
+        setClientSideFilter(!forceServerSideFiltering
+                && dataCommunicator.getItemCount() <= getPageSizeDouble()
+                // We do want to perform server side filtering (send client
+                // filter to the server) if there is at least one item count
+                // change listener added, even in case of in-memory data
+                // provider and even if the item count < page size.
+                // This allows us be always notified about size change.
+                && itemCountListeners.get() == 0);
```
pleku, line 977: As commented about this, it feels odd behavior change, that could be avoided with just delegating the client side filtering result count to server side instead

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -1160,34 +1563,29 @@ private void reset() {
...
     }
 
-    /**
-     * Callback for Data Communicator lazy initialization
-     */
-    @FunctionalInterface
-    private interface DataCommunicatorInitializer extends Serializable {
-        void init();
+    private void enableDataProviderFetch() {
+        dataCommunicator.setFetchDisabled(false);
```
pleku, line 1573: Oh damn my bad review, just now realized that there is a double negation which is against our conventions 😞 
Should be `fetchEnabled` instead

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -711,9 +1082,41 @@ public void setDataProvider(ItemFilter<T> itemFilter,
         setDataProvider(listDataProvider,
                 filterText -> item -> itemFilter.test(item, filterText));
 
-        // Force the data communicator initialization eagerly because in-memory
-        // data is used (ListDataProvider)
-        initDataCommunicator();
+        // Enable fetch and size queries eagerly because in-memory
+        // data is used and getting the collection size is not so
+        // heavy like a database call
+        enableDataProviderFetch();
```
pleku, line 1088: This does eagerly something that was postponed in the call a few lines earlier - which is a code smell.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -1160,34 +1563,29 @@ private void reset() {
...
-             */
-            dataCommunicatorInitializer.init();
-        } else if (dataCommunicator == null) {
-            /*
-             * If the user hasn't provided any data, initialize with empty data
-             * set.
-             */
-            setItems();
+    private void initDataCommunicatorIfEmpty() {
+        if (dataCommunicator == null) {
```
pleku, line 1567: Is there any point in delaying this instead of doing it directly in the constructor and removing this method and the places where it is called from (getting data view) ??? If there is, please answer as comment in code

pleku (changes requested):

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -254,29 +260,107 @@ private void storingCustomValues() {
...
+         * items and as the user scrolls down it automatically increases the
+         * size by 200 until the backend runs out of items.
+         *
+         * Both the estimated item count and its increase can be customized to
+         * allow the user to scroll down faster when the backend will have a lot
+         * of items.
          */
-        comboBox.setDataProvider(service::fetch, service::count);
+        lazyDataView.setItemCountEstimate(1000);
+        lazyDataView.setItemCountEstimateIncrease(1000);
```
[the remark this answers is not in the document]
mshabarov: I would keep this paragraph "By default, the combo box will initially adjust the scrollbar to 200...".

Increase of 100 is less than the default one of 200, and this may confuse the users, because the showcase of this demo is a fast scrolling and the question then would be "How 100 can be faster than 200", even the page size is decreased comparing to default one.

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -254,29 +260,107 @@ private void storingCustomValues() {
...
         PersonService service = new PersonService();
         /*
-         * This data provider doesn't load all the items to the server memory
-         * right away. The component calls the first provided callback to fetch
-         * items from the given range with the given filter. The second callback
-         * should provide the number of items that match the query.
+         * By using these callbacks the ComboBox doesn't load all the items to
+         * the server memory right away. The ComboBox calls the first provided
+         * callback to fetch items from the given range with the given filter.
+         * The second callback should provide the number of items that match
```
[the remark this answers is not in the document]
mshabarov: done

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -286,14 +370,15 @@ private void pagedRepository() {
          * For those backend repositories which use paged data fetching, it
          * is possible to get the page number and page size from Query API.
          */
-        comboBox.setDataProvider(DataProvider.fromFilteringCallbacks(
-                query -> service.fetchPage(query.getFilter().orElse(""),
+        comboBox.setItems(
+                query -> service.fetchPage(query.getFilter().orElse(null),
                         query.getPage(), query.getPageSize()),
-                query -> service.count(query.getFilter().orElse(""))));
+                query -> service.count(query.getFilter().orElse(null)));
```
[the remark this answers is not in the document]
mshabarov: Good point! Count query is not a focus for this demo. Removed

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxListDataViewIT.java
```diff
@@ -0,0 +1,174 @@
...
+
+        verifySelectedPerson(6);
+
+        showNextPerson();
+
+        verifySelectedPerson(7);
+
+        showPreviousPerson();
+
+        verifySelectedPerson(5);
```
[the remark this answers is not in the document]
mshabarov: The code on the page is made so that the previous item considered relatively to 6, not 7. I will modify it for better readability

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxListDataViewIT.java
```diff
@@ -0,0 +1,174 @@
...
+
+        // Filter by Age (programmatic filter)
+        setAgeFilter("50");
```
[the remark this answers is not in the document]
mshabarov: Yes, that's ugly. Will split it out

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxDataViewTest.java
```diff
@@ -0,0 +1,309 @@
...
+        dataView.addItemCountChangeListener(event -> {});
+
+        // Trigger ComboBox::refreshAllData()
+        dataProvider.refreshAll();
+
+        Assert.assertFalse(Boolean.parseBoolean(getClientSideFilter()));
+    }
+
+    @Test
+    public void addItemCountChangeListener_removeListeners_serverSideFilterDisabled() {
```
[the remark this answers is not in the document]
mshabarov: ComboBox has a feature to not call the server side for fetching the items if the items count < page size:

```
setClientSideFilter(!forceServerSideFiltering && getDataProvider()
                .size(new Query<>()) <= getPageSizeDouble());
```

In that case the filtering is done on client side only, that means `flush()` method is never invoked = listeners are never notified. But if the developer adds an item count listener, then it's good to notify these listeners anyway, even if < page Size.

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxListDataViewTest.java
mshabarov: Good catch, will improve that

mshabarov:
> It has been too long time since the initial review - did we sort this out somehow ?
Why did we choose to have F type instead of String which the CB always provides and then force the conversion to happen in the callback implementation instead ?

Because I use this callback class instead of `FetchCallback` in `setItems(FetchItemsCallback<T, C> fetchCallback, SerializableFunction<String, C> filterConverter)`.

If I revert it to `FetchCallback` then I will experience ambiguous method call:

```
Ambiguous method call. Both setItems(FetchCallback<Person, Integer>, SerializableFunction<String, Integer>)
in ComboBox and
setItems(FetchCallback<Person, String>,CountCallback<Person, String>)
in HasLazyDataView match
```

in the following use case:

```Java
comboBox.setItems(
        query -> service.fetch(query.getFilter().orElse(null),
                query.getOffset(), query.getLimit()),
        query -> service.count(query.getFilter().orElse(null)));
```

No other reason for changing `String` to `F`

mshabarov:
> Is this now fixed ?

Not yet, I've only created a ticket for that

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountEstimateIncreaseComboBoxIT.java
```diff
@@ -0,0 +1,101 @@
...
+        customIncrease = 500;
+        setEstimateIncrease(customIncrease);
+
+        scrollToItem(comboBoxElement, newCount - 10);
+
+        verifyItemsSize(newCount + customIncrease);
+    }
+
+    @Test
+    public void customIncrease_reachesEndBeforeEstimate_sizeChanges() {
```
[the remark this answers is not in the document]
mshabarov: done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: This is done to provide a use case, when you call `dataView.setItemCountEstimate(42)` and want to see the scroller size change immediately after open the dropdown. Even if the assumed size changed, the requested range is still [0, 50] and size changes are not committed to client. Grid has this line of code also in the `UpdateQueue`. Comment to a code is not probably reflect the meaning of this change

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Oh, yes, you're right, no need to postpone it in the constructor. Will fix that

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Yes, looks ugly. I will add the check depending on DataProvider type

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Thanks! Will change the method name

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Hmm, some leftover from intermediate version. I propose to add this listener only when the backend data provider is assigned and remove this registration as soon as event is triggered

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: > Isn't client side filtering used only with in-memory data ?

Yes, if `ListDataProvider` or generic `InMemoryDataProvider` is used and items count < page size, then client filter works without requests to server. Basically this is how it works before data view changes. I've added it for `InMemoryDataProvider` and forced server requests in case of listeners.

mshabarov: > whole filtering mode would change when the only reason why this is necessary is that the server side needs to know the amount of items.

Mayhaps, but IMO more odd DX will be expect item count change triggering, but getting nothing because of client filter mode.

Well, when I explored it, it took some time for me to figure out why I didn't get those terrible event...oh, because of only 7 items :)

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
```diff
@@ -283,11 +283,12 @@ public void setValueNull_selectedItemNull() {
...
     @Test
-    public void setValueWithoutItems_throw() {
-        expectIllegalStateException(
-                "Cannot set a value for a ComboBox without items.");
+    public void setValueWithoutItems_valueIsSet() {
         ComboBox<String> combo = new ComboBox<>();
         combo.setValue("foo");
+        // Since the combo box is always initialised with data communicator on
+        // instantiation phase, setting the value is possible
+        Assert.assertEquals("foo", combo.getValue());
```
pleku, line 291: How can a combo box without any items have a selected item as its value ??? What does that mean ? When would it be useful instead of just a developers error ? Do we support that for other selection components too ?

It feels to me that you accidentally enabled this by internal changes and it should not still be possible

pleku: The related feature for this is allowing new items and I don't think this ^ is the way to do it

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
```diff
@@ -283,11 +283,12 @@ public void setValueNull_selectedItemNull() {
     }
```
pleku, line 284: Are the changes to `setDataProvider` regarding the data communicator initialization and `enableFetch` covered by unit tests ? At least not in this class

pleku (changes requested):
Couple of more comments and it seems that some previous comments have been addressed but some are not.

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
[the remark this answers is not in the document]
pleku: Not done

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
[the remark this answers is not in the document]
pleku: You missed the point - this is not grid and since it has filtering that the user should use the expectancy is that there will be fewer items and less scrolling so the demo would be instead to showcase the customizability in a use case, that I think, is more natural for CB.

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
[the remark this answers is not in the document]
pleku: Not done ???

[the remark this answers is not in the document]
mshabarov: The code diff is outdated.
I've removed the count query from the paged repos demo:
```
        comboBox.setItems(
                query -> service.fetchPage(query.getFilter().orElse(null),
                        query.getPage(), query.getPageSize()));
```

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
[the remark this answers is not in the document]
mshabarov: Code snippet above is outdated for some reason.
The wording is changed actually:
```
         * The second callback provides the number of items that match the
         * query.
```

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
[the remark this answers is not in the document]
mshabarov: Got it, thanks! So I modified it as follows:
```
        /*
         * By default, the combo box will initially adjust the scrollbar to 200
         * items and as the user scrolls down it automatically increases the
         * size by 200 until the backend runs out of items.
         *
         * In case of huge number of items in the backend, combo box's text
         * filter can be used to decrease the number of shown items and
         * you can customize the estimated item count and its increase to allow
         * the user scroll through the filtered items set.
         */
        comboBox.setPageSize(20);
        lazyDataView.setItemCountEstimate(100);
        lazyDataView.setItemCountEstimateIncrease(100);
```

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
mshabarov: I thought that since the combo box now has a `dataCommunicator` instance being created in the CTOR, we could remove  `if (dataCommunicator == null)` code block. And the old `setValue()` allowed new values on data communicator with no items, but not null. However, you're right, there is no sense to set value to empty combo box. So, I propose: 1. Revert the removed code block, since `setValue` still can be called before DC init accidentally  2. Add the items count check.

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
mshabarov: No, they were missed, thanks. `LazyLoadingIT` contains the test for verifying that the data provider is not queried eagerly if of backend type, but it's not enough. I've added the unit tests.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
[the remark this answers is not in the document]
mshabarov: These changes have been reverted and `setItems` methods which take the filter converter renamed to `setItemsWithFilterConverter`.

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/AbstractItemCountComboBoxIT.java
mshabarov: Yes, fixed

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxListDataViewIT.java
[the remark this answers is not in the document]
mshabarov: Done

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxListDataViewTest.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: This code has been removed and item count notification now goes from client to server side in client filter mode.

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
```diff
@@ -455,6 +460,32 @@ private void filteringAndSortingWithDataView() {
...
+                "Orange", "Banana", "Kiwifruit", "Grapefruit");
+
+        Span fruitCountSpan = new Span(
+                "Fruits count: " + dataView.getItemCount());
+
+        // Client filter usually changes the displayed items count in combo
+        // box. With data view, one can subscribe to combo box's items count
+        // change events and get a notifications on server side.
+        dataView.addItemCountChangeListener(event -> fruitCountSpan
+                .setText("Fruits count: " + event.getItemCount()));
```
pleku, line 479: I don't think this "valid" - it is not the fruit count but that the "matches count" when filtering. And that was what I meant with this example, that it would just show the number when filtered. And that number doesn't make any sense here with just so few items.

Maybe the idea for the demo was not that good after all - or at least like this it is just confusing. Not sure if it would just be about showing e.g. `(123)`  (all items) first and then when filtering it would show `(32/123)` or just, but I realize that would actually require even more effort ...

And now as I see that this demo doesn't even make sense, I'm starting to question that why would the server side item count event be fired when the component internal filtering is considered and the number of items doesn't change ? Because the semantics of the item count event (for all other components) are that it is only fired when the item count changes based on external filtering. I should have probably taken more thought about this even when I saw the confusing switch to server side filtering based on whether or not the listener is added.

One way to look at this would be to say that in the combo box case the item count change is never fired for internal filtering changes - since after the filtering is done and the user selects an item, the count reverts back to what it was anyway. So basically the real item count doesn't even change, just how much the user sees at the moment has changed. If anyone would need the number of items the CB is showing, it should be another event `FilteredItemCount` or such, but I don't know if that is even ever needed!

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -868,13 +827,26 @@ public void setDataProvider(DataProvider<T, String> dataProvider) {
             userProvidedFilter = UserProvidedFilter.YES;
         }
 
+        // Fetch from data provider is enabled eagerly if the data provider
+        // is of in-memory type and it's not empty (no need to fetch from
+        // empty data provider). Otherwise, the fetch will be postponed until
+        // dropdown open event
+        final boolean enableFetch = InMemoryDataProvider.class
```
pleku, line 834: What about `DataProvider::isInMemory` ??? If one has a custom in-memory data provider that is not based on `InMemoryDataProvider` ?

pleku (changes requested):
Close but no dice yet

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
mshabarov: Yes, that makes sense! I will revert these changes then, and make the data communicator to recognise whether it should fire the event or not.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Good catch! Thanks! Will fix

vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
mshabarov: Done.

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxDataViewTest.java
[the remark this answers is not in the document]
mshabarov: This code has been removed. Another issue is that the client filter change should not trigger item count change event on the server. But this will be a separate ticket.

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/dataview/ComboBoxListDataViewTest.java
mshabarov: This code has been removed.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: This code has been removed. Another issue is that the client filter change should not trigger item count change event on the server. But this will be a separate ticket.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: > item count notification now goes from client to server side in client filter mode.

It has been removed too. Client filter change should not trigger item count change event on the server. But this will be a separate ticket.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

mshabarov:
> > Is this now fixed ?
> 
> Not yet, I've only created a ticket for that

Fixed

vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
mshabarov: Done

vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java
mshabarov: Disabling the firing an item count change event on the server side while applying a client filter is not a trivial change. This is closely connected with `ListDataView` API (add/remove/getItems/addFilter). The main problem is how to simultaneously disable event firing for client filter and enable for data view API. This will covered by a separate ticket.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Done

mshabarov:
This is a ticket for item count change event issue https://github.com/vaadin/vaadin-flow-components/issues/210

pleku (approved):

PR #394, opened 2020-09-21
Don't set value if Element has property "value"

### LINKED ISSUE #393: GeneratedVaadinComboBox CTOR is unaware of possibly already existing "value" property in underlying Element
This is failing test:

```
private class ComboBoxWithInitialValue
            extends GeneratedVaadinComboBox<ComboBoxWithInitialValue, String> {
        ComboBoxWithInitialValue() {
            super("", null, String.class, value -> value, value -> value);
        }
    }

@Test
    public void elementHasValue_wrapIntoField_propertyIsNotSetToInitialValue() {
        Element element = new Element("vaadin-combo-box");
        element.setProperty("value", "foo");
        UI ui = new UI();
        UI.setCurrent(ui);
        VaadinSession session = Mockito.mock(VaadinSession.class);
        ui.getInternals().setSession(session);
        VaadinService service = Mockito.mock(VaadinService.class);
        Mockito.when(session.getService()).thenReturn(service);

        Instantiator instantiator = Mockito.mock(Instantiator.class);

        Mockito.when(service.getInstantiator()).thenReturn(instantiator);

        Mockito.when(
                instantiator.createComponent(ComboBoxWithInitialValue.class))
                .thenAnswer(invocation -> new ComboBoxWithInitialValue());
        ComboBoxWithInitialValue field = Component.from(element,
                ComboBoxWithInitialValue.class);
        Assert.assertEquals("foo", field.getElement().getPropertyRaw("value"));
    }
```

denis-anisimov:
Fixes #393

### DIFF
--- a/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
+++ b/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
@@ -393,7 +393,7 @@ public abstract class GeneratedVaadinComboBox<R extends GeneratedVaadinComboBox<
      * the returned value may not be the same as in client side.
      * </p>
      * 
-     * @return the {@code size} property fromw the webcomponent
+     * @return the {@code size} property from the webcomponent
      */
     protected double getSizeDouble() {
         return getElement().getProperty("size", 0.0);
--- a/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
+++ b/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
@@ -393,7 +393,7 @@ public abstract class GeneratedVaadinComboBox<R extends GeneratedVaadinComboBox<
      * the returned value may not be the same as in client side.
      * </p>
      * 
-     * @return the {@code size} property from the webcomponent
+     * @return the {@code size} property fromw the webcomponent
      */
     protected double getSizeDouble() {
         return getElement().getProperty("size", 0.0);
@@ -1424,7 +1424,7 @@ public abstract class GeneratedVaadinComboBox<R extends GeneratedVaadinComboBox<
             SerializableFunction<T, P> modelToPresentation) {
         super("value", defaultValue, elementPropertyType, presentationToModel,
                 modelToPresentation);
-        if (getElement().getProperty("value") == null && initialValue != null) {
+        if (initialValue != null) {
             setPresentationValue(initialValue);
         }
     }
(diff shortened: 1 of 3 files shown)


vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
```diff
@@ -1466,21 +1464,54 @@ public GeneratedVaadinComboBox(T initialValue, T defaultValue,
...
+     *            returns a property value
+     * @param <P>
+     *            the property type
+     */
+    public <P> GeneratedVaadinComboBox(T initialValue, T defaultValue,
+            Class<P> elementPropertyType,
+            SerializableBiFunction<R, P, T> presentationToModel,
+            SerializableBiFunction<R, T, P> modelToPresentation) {
+        this(initialValue, defaultValue, elementPropertyType,
+                presentationToModel, modelToPresentation, false);
```
mshabarov, line 1512: Why do we use `false` (mandatory value) here? Isn't it better to allow `value` property by default for `GeneratedVaadinComboBox` child classes?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
```diff
@@ -1445,7 +1444,6 @@ public GeneratedVaadinComboBox(T initialValue, T defaultValue,
             boolean acceptNullValues) {
         super("value", defaultValue, acceptNullValues);
         if (initialValue != null) {
-            setModelValue(initialValue, false);
```
mshabarov, line 1448: Why don't we need `setModelValue ` anymore? How otherwise the model value will be set?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
```diff
@@ -393,7 +393,7 @@ protected void setPageSize(double pageSize) {
      * the returned value may not be the same as in client side.
      * </p>
      * 
-     * @return the {@code size} property from the webcomponent
+     * @return the {@code size} property fromw the webcomponent
```
mshabarov, line 396:
```suggestion
     * @return the {@code size} property from the webcomponent
```

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
mshabarov: Is it initialised automatically in other place based on `value` property of the web component?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
mshabarov: I assume `false` is used when the initial value is present as a constructor argument, right?

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
denis-anisimov: The opposite question: why it should be used ?
And this is in fact a good question: why use extra method if everything works without it ?

There is no any failure in any IT/unit tests.
Which proves perfectly that this is extra method.

`setPresentationValue` updates both values: model is updated as a consequence of presentation value change.

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
denis-anisimov: This CTOR is preserved because of backward compatibility.
It has `initialValue`. If I pass an initial value then I expect that it will be set.
The CTOR which allows to use initial value optionally has a flag which has to be specified explicitly: the user should make the exact decision whether the initial value is optional or not.
By default initial value has to be set.

mshabarov: Yes, thanks, now it's clear

vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
mshabarov: Yes, you're right, it updated in model as follows:
```
 setter.setElementValue(element, propertyName, presentationValue);
```
in `AbstractSinglePropertyField`. Then it's excessive.

mshabarov (approved):

